PR #5, opened 2017-03-18
Bump serde dependencies

warmwaffles:
It looks like a lot has changed since 0.8.*

Simply bumping the dependency made the entire library explode with errors about `serde_json::Map<String, Accessor>` etc... not being defined. Instead I opted to use `std::collections::HashMap` to represent the key value store.

I'm not sure if this entirely the right solution, but it seems to compile fine and the example still works.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,12 +8,12 @@ repository = "https://github.com/Alteous/gltf"
 readme = "README.md"
 keywords = ["glTF", "3D", "asset", "model", "scene"]
 license = "MIT"
- 
+
 [dependencies]
 gl = "0.6"
-serde = "0.8"
-serde_derive = "0.8"
-serde_json = "0.8"
+serde = "0.9"
+serde_derive = "0.9"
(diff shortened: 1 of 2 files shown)


warmwaffles:
Looks like it is failing on nightly

alteous:
The serde_json team decided to make `Map` an [opaque type](https://github.com/serde-rs/json/commit/262e144c0ebccc9f85a26b17d9e07b8e95eeed04) for the 0.9 release. Swapping it for our own `Map` type is a sensible proposal, but I would prefer to keep the same name. That is, to replace

```rust
pub use serde_json::value::{Map, Value};
```
with
```rust
pub use std::collections::HashMap as Map;
pub use serde_json::Value;
```

warmwaffles:
Heh, your new branch is 5 commits behind master and has merge conflicts. https://github.com/alteous/gltf/compare/v0.4.0

alteous:
My bad, you should be good to go now.

PR #7, opened 2017-03-18
Allow nightly and beta channels to fail

warmwaffles:
Since they aren't stable it's probably best to let them fail. Found this out when trying to update the serde dependency.


warmwaffles:
Looks like it was just a hiccup with the nightly build from the previous day.

alteous:
Thanks, I've merged this manually into branch v0.4.0.

PR #9, opened 2017-03-22
Reimplemented glTF tree

warmwaffles:
* Added license header to new files
* Added gltf_display example
* Added examples to README

Based on the conversations had in https://github.com/alteous/gltf/issues/4

We can make this a work in progress thing.


alteous:
Great, thanks!

PR #10, opened 2017-03-23
Add enum string to accessors and animations

warmwaffles:


src/v1/buffer.rs
```diff
@@ -6,12 +6,12 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-enum_number!(
+enum_number! {
     ViewType {
```
alteous: Target would be a better name

src/v1/texture.rs
```diff
@@ -15,64 +15,64 @@ enum_number!(
...
 
 impl Default for Target {
     fn default() -> Target {
         Target::Texture2d
     }
 }
 
-enum_number!(
+enum_number! {
     TexelType {
```
alteous: I think the following is more terse, self-documenting, and 'rustic':
```rust
TexelType {
    U8 = 5121,
    U16_R5_G6_B5 = 33635,
    U16_R4_G4_B4_A4 = 32819,
    U16_R5_G5_B5_A1 = 32820,
}
```
Let's consider changing this in the next release.

src/v1/camera.rs
```diff
@@ -24,7 +37,7 @@ pub struct Perspective {
     /// The floating-point distance to the near clipping plane.
     ///
     /// `zfar` must be greater than `znear`.
-    pub znear: f32, 
+    pub znear: f32,
```
alteous: I prefer `z_near` etc but that's really just a matter of style. I think 'proper' snake_case actually mandates no single letter words at the start (hence why standard library has `btree_map` and not `b_tree_map`) but in this particular case `z_near` just seems so much easier on the eye than `znear`.

src/v1/animation.rs
```diff
@@ -8,26 +8,47 @@
 
 use std::collections::HashMap;
 
-#[derive(Debug, Default, Deserialize, Serialize)]
+enum_string! {
```
alteous: `Path` would be more terse.

src/v1/accessor.rs
```diff
@@ -6,7 +6,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-enum_number!(
+enum_number! {
     ComponentType {
```
alteous: This could be more 'rustic' with the following:
```
ComponentType {
    I8 = 5120,
    U8 = 5121,
    F32 = 5126,
    F64 = 5127,
}
```
This would need to be discussed further because it conflicts with `Material` (`FloatVec3` etc). If desired we'll make the change in a future release.

src/v1/texture.rs
warmwaffles: I thought enums weren't supposed to be all upper case like that

alteous: They're not really all upper case. It just looks that way because every 'word' only has one letter. The _ pattern is definitely unconventional, but it conveys meaning much easier. Compare this to [vulkano](https://github.com/tomaka/vulkano/blob/master/vulkano/src/format.rs#L364) and [glium](https://github.com/tomaka/glium/blob/master/src/image_format.rs#L74) and hopefully you'll see what I mean.

alteous: I guess the conventional way would be like this:
```rust
TexelType {
    U8 = 5121,
    U16R5G6B5 = 33635,
    U16R4G4B4A4 = 32819,
    U16R5G5B5A1 = 32820,
}
```
Which would you prefer?

alteous: [gfx](https://github.com/gfx-rs/gfx/blob/master/src/core/src/format.rs#L339) takes a similar approach.

alteous: Another alternative might be
```rust
TexelType {
    UnsignedByte = 5121,
    UnsignedShortR5G6B5 = 33635,
    UnsignedShortR4G4B4A4 = 32819,
    UnsignedShortR5G5B5A1 = 32820,
 }
```
It's quite awkward that `UnsignedByte` is by itself when really that means `UnsignedByteR8B8G8A8`. On second thoughts (and to avoid the [bikeshed effect](https://en.wikipedia.org/wiki/Law_of_triviality)) I think your proposal will be best.

warmwaffles: I think if we try to keep inline with gfx and volkano and glium we should be fine since those three libs are most likely to be used in conjunction with this library.

src/v1/animation.rs
warmwaffles: Hmm, I was trying to avoid conflating with `std::Path`

warmwaffles:
Changes made

warmwaffles:
Also curious @alteous do you have an example app that uses `gltf v0.3.0`? I am working on a basic model preview tool to do a sanity check that the assets are loaded.

alteous:
Great job, thanks!

alteous:
I have an abandoned renderer that used an unpublished gltf version between 0.2.1 and 0.3.0. I've managed to dig it out of the grave but a few tweaks are necessary to get it back up to date. Would you like to see the code in its current state?

alteous:
The renderer worked correctly prior to abandonment, I should add.

warmwaffles:
> Would you like to see the code in its current state?

Yes

alteous:
Okay, here you go: https://github.com/alteous/gltf-previewer

I'd be more than happy to discuss the project in detail if you have any queries about the implementation or if you want to continue the development with me.

warmwaffles:
Perfect, I'll check it out and poke around.

Mind you I am still learning rust so I may end up implementing stuff in funky non-normal ways.

alteous:
@warmwaffles If you're planning to write your own glTF renderer, I would feel bad not telling you about the mistakes I learned the hard way when writing `gltf-previewer`. Please don't feel as if you have to take any notice of these, but in the interest of potentially saving you a lot of wasted effort...

- Use `nalgebra` as your linear algebra library. Most people use `cgmath`, but I've found it very annoying to use. You'll be constantly importing meaningless traits like `Matrix`, `SquareMatrix`, `EuclideanSpace` and making dumb `Point -> Vec -> Point` conversions all over the place. Also, `nalgebra` has inter-operable sister libraries like `nphysics` which you might find useful at a later date.
- For rendering glTF, you will _probably_ have to write your own back-end. I'd be interested to know if you have any success with third-party back-ends otherwise. For `gltf-previewer` I had previously tried `glitter`, `glium`, and `gfx`. `glitter` was most appropriate for the use case but I couldn't understand how to use it. `glium` and `gfx` on the other hand were much easier to use but lacked the low-level features required to accommodate the flexibility of the glTF format. Eventually I wrote my own back-end ([`yagl`](https://github.com/alteous/gltf-previewer/tree/master/yagl)) which was by far a much smoother experience.
- For window management, `glfw` and `winit` are more reliable alternatives to `glutin`. I say this because creating an OpenGL context with `glutin` on Linux has been troublesome for quite a while.

Incidentally, the reason why I abandoned  `gltf-previewer` was not because of the difficulty of the implementation, but rather I decided that the mandatory, explicit shaders in glTF 1.0 were unsuitable for my personal project. Hence I was very pleased when you informed me of the new specification!

warmwaffles:
> explicit shaders in glTF 1.0 were unsuitable for my personal project

Yea that was my deal breaker when I was writing a demo in C to render gltf models. What I was wanting from glTF is an open standard for 3d models that was efficient and supported animations as well.

As for `cgmath` that was my findings as well. I tinkered with `nalgebra` and it seems the closest to c++'s `glm` library.

PR #16, opened 2017-05-06
Remove AsRef from open to remove the code bloat

MaikKlein:
See https://github.com/rust-lang/rust/issues/41782

I guess we could also make the generic `open` function a free standing function.

### DIFF
--- a/src/v1/mod.rs
+++ b/src/v1/mod.rs
@@ -184,7 +184,7 @@ pub struct Gltf {
 }
 
 impl Gltf {
-    pub fn open<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
+    pub fn open(path: &Path) -> Result<Self, Error> {
         let mut file = File::open(path)?;
         let mut json = String::new();
         file.read_to_string(&mut json)?;


bluss:
I'd recommend using the solution in this PR https://github.com/PistonDevelopers/image/pull/518

alteous:
Thanks for reporting the long build time. This is clearly getting out of hand.

@bluss ' suggestion would be a non-breaking change, so I'll go ahead and make that change and see if the build times improve.

It's worth noting that this crate is going through a lot of changes since serde 1.0 has been released and that the glTF 2.0 specification will be released very soon. So in the near future build times could be even longer. Any suggestions to prevent this would be greatly appreciated.

alteous:
I've commited these changes to [`patches`](https://github.com/alteous/gltf/commit/6b2c8642031504df6aaad03f99e7911feb93ae5d) but it seems to have made build times even worse! See my comment in the commit for some times.

bluss:
Note here:

Generic items are not compiled to machine code. They need to be instantiated completely with fixed types for each type parameter ("monomorphization") before they are compiled. It typically happens that the library crate just type checks its generic items and just puts them as metadata items into the library crate.

The de-generic solution has the following effects:

1. The same code (open_impl) is only compiled one time instead of once per `AsRef<Path>` variation
2. The same code (open_impl) is only compiled one time inside the gltf crate, instead of once for each crate that uses and instantiates the code

Note that (2) Means that open_impl increases the number of times that code is increased by one. It was generic and never codegenned as part of compiling gltf before. The benefit is that it doesn't have to be compiled or recompiled in the crate where it is used.

It is thus expected that gltf's compile time increases here, but the usage site's crate compile time decreases.

bluss:
Let's make scenarios and what I expect to happen.

We have a project's crate A and it depends on and uses gltf as a dependency.

1. Old code, A uses `Gltf::open` once.
  - Fresh compile effect: Long compile, but most of it in A, not in gltf.
  - Recompile effect: Still long, recompile of A means recompiling all of Gltf::open again.
2. New code, A uses `Gltf::open` once.
  - Fresh compile effect: Nothing, the same code is compiled, but in different place now than before
  - Recompile effect: Recompile of A is quick, `Gltf::open` was compiled as part of `gltf.
3. Old code, A uses `Gltf::open` twice, once with `&str` and once with `&Path`.
  - Fresh compile effect: Even longer compile. `Gltf::open` is compiled from scratch twice.
  - Recompile effect: Even longer compile. `Gltf::open` is compiled from scratch twice.
4. New code, A uses `Gltf::open` twice, once with `&str` and once with `&Path`.
  - Fresh compile effect: Long compile, but only compiles the meat of `Gltf::open` once.
  - Recompile effect: Recompile of A is quick, `Gltf::open` was compiled as part of `gltf.

bluss:
Here's are stock conclusions:

- If the long compile time of `gltf` is bothering you, keep the de-generization solution to reduce generics bloat, but mark the inner function `#[inline]`. Then you are again punting the code generation to the usage site
- If `Gltf::open` is just a rarely used feature of the library, mark it `#[inline]` and let users compile it on demand

alteous:
Great explanation. I understand how this will affect the build times from the user's perspective now.

I think it's more important for the build times to be reasonable for the user, so I'll make the merge and publish a new minor version once the tests pass. `Gltf::open` is supposed to be the entry point in this version of the crate, so I won't be marking it `#[inline]` anytime soon. 

#### Side note

In the next release `Gltf::open` will be replaced with functions `gltf::v1::import` and `gltf::v2::import` for glTF versions 1.0 and 2.0 respectively. I'll make sure the same build time optimisation is implemented.

MaikKlein:
Thanks, I guess I can close this PR now.

Issue #20, opened 2017-05-16
Fix loading glTF 2.0 sample models


alteous:
@bwasty quite rightly reported that `tests/import_v2.rs` is failing on the latest 2.0 sample models.

On a side note, it would be worth rewriting the test to read the directory entries.

bwasty:
Got another fix in progress (https://github.com/alteous/gltf/compare/incoming...bwasty:fix_box_loading) and found one issue was actually in the sample models: https://github.com/KhronosGroup/glTF-Sample-Models/issues/68.

bwasty:
Done. I agree that `it would be worth rewriting the test to read the directory entries`, but for now I just wanted to quickly check if there are more issues, so I did it manually.

alteous:
Thanks for fixing this.

FYI I'm currently rewriting the tests.

PR #23, opened 2017-05-20
Make texture sampler optional

bwasty:
Partial fix for #20:
The tests failed on `Corso.gltf` because it doesn't have `sampler` attributes on its `textures`. According to the spec, the field is not required: https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#texture


alteous:
Thank you!

PR #30, opened 2017-06-20
Minor improvement to Wrapper branch

fst3a:
This PR brings tiny improvement: it uses serde_json::from_reader instead of from_slice.  The advantage is that it does not need to load raw glTF data anymore but rather lets serde_json read and parse as it needs.  It brings tiny memory usage improvement since serde_json parses char by char and reads as it needs and never keeps whole raw thing in memory.  This removes Importer::gltf field as it is deemed redundant and unneeded.

Additionally, this PR changes to use all upcoming minor versions of 1.x.x of serde and brings rustfmt.toml which should be mostly compatible with the style that is already used in project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,9 @@ include = [
 ]
 
 [dependencies]
-serde = "1.0"
-serde_derive = "1.0"
-serde_json = "1.0"
+serde = "1"
+serde_derive = "1"
+serde_json = "1"
 gltf_derive = { path = "./gltf_derive" }
 
 [features]
@@ -27,4 +27,3 @@ extras = []
 [[example]]
 name = "gltf_display"
 path = "examples/display/main.rs"
-
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,12 @@
+chain_one_line_max = 80
+closure_block_indent_threshold = -1
+condense_wildcard_suffixes = true
+error_on_line_overflow = false
+max_width = 80
+reorder_imported_names = true
+reorder_imports = true
+reorder_imports_in_group = true
+take_source_hints = true
+where_style = "Legacy"
+wrap_comments = true
+fn_args_layout = "Visual"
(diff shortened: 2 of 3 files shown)


alteous:
Thanks for the pull request. It's great to see new contributors.

I'm not opposed to this change but it does raise some questions in regards to the library design. Below are the outlined concerns:

This may seem like a 'minor' improvement at first glance but actually there is a remarkable difference between `serde_json::from_slice` and `serde_json::from_reader`, namely the former can take advantage of so-called 'zero-copy' deserialisation whereas the latter can not.

The primary advantage of zero-copy deserialisation is that it avoids making copies of strings (of which there are many) since slices are taken from the input buffer data instead. This isn't well documented but there is some useful information in the [release notes](https://github.com/serde-rs/serde/releases/tag/v1.0.0) - scroll down a bit. It is a fair point that loading the whole glTF JSON and calling `from_slice` may be unnecessary, but I would argue the data is contiguous and minimises the number of heap allocations - `from_reader` does the opposite! Knowing that the library is making lots of heap allocations in a multi-threaded user application makes me feel guilty, because every call to `(je)malloc` takes a mutex lock.

The reason why the `gltf` field exists in `Importer `is to facilitate zero-copy deserialisation, as we have to hand out references to owned data. There is an extra advantage in this strategy in that `gltf` needs to be (re-)allocated between import calls only when absolutely necessary.

I think it would be reasonable to let the user decide what flavour of deserialisation they prefer by providing two kinds of `Importer`: `Owned` (the version in this PR), and `ZeroCopy` (the current version).

I'll plan to make this change after merging the PR, but beforehand I would like to hear your feedback in case there's anything else worth considering. It would be nice to hear @bwasty 's opinion on this too.

fst3a:
Actually, ignore change from_slice to from_reader, i'm stupid.  I just benchmarked (in debug) and it seems that from_slice gives much greater performance both in multithread and singlethread workloads.  Additionally, I tried to benchmark in --release, but it seems that it takes enormous amount of time to build in release (on stable toolchain), is this known or should issue be created for this one?

fst3a:
Well at least rustfmt should be useful change, lol.

alteous:
> i'm stupid

Don't say that, it's not true! We appreciate any feedback and constructive criticism.

> Well at least rustfmt should be useful change, lol.

Yes, it would be. I think the new style from rustfmt (coming soon) is much easier on the eyes.

PR #41, opened 2017-06-30
Added validation::Action

alteous:
This may be used to exit early from validation.


alteous:
In this scenario, `git` stands for goddamn idiotic truckload...

PR #42, opened 2017-06-30
Allow the validation process to end early

alteous:
This is achieved using a new type, namely `Action`, which can take the value `Stop` or `Continue`. The validation callback decides whether the continue the validation or not.

### DIFF
--- a/gltf_derive/src/lib.rs
+++ b/gltf_derive/src/lib.rs
@@ -38,20 +38,28 @@ fn expand(ast: &syn::MacroInput) -> quote::Tokens {
     let validations: Vec<quote::Tokens> = fields.iter()
         .map(|f| f.ident.as_ref().unwrap())
         .map(|ident| {
-                 use inflections::Inflect;
-                 let field = ident.as_ref().to_camel_case();
-                 quote!(self.#ident.validate(root, || path().field(#field), report))
-             })
+            use inflections::Inflect;
+            let field = ident.as_ref().to_camel_case();
+            quote!(
+                if {
+                    self.#ident.validate(root, || path().field(#field), report)
+                } == ::validation::Action::Stop {
+                    return ::validation::Action::Stop;
+                }
+            )
+        })
         .collect();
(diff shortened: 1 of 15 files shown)


alteous:
I'm not actually sure how long it takes; let's benchmark and see.

It may not be important to terminate early, but at least we have the option. Hopefully the compiler is smart enough to optimise away the `if action == Stop { return Stop; }` checks if the callback always returns `Continue`.

Now, thinking back to [Mike Action's infamous data oriented design presentation](https://www.youtube.com/watch?v=rX0ItVEVjHc), I imagine the most probable case is that the glTF is perfectly OK. In this scenario, it may be better to validate everything anyway in order to significantly reduce the number of branches.

Let me investigate some more and get back to you.

alteous:
Thinking back to your example in https://github.com/alteous/gltf/issues/13#issuecomment-312188925, specifically the line `.validation(Validation::Complete) // or Minimal`, perhaps there is a future optimisation to be made:

```rust
/// Validates glTF data.
trait Validate<'a> {
    /// Validate only the essential invariants required by the crate to function properly.
    ///
    /// This includes but is not limited to:
    /// * Index bounds checking
    /// * Camera types
    fn minimally<P, R>(&self, root: &Root<'a>, path: P, report: &mut R) -> Action
    where
        P: Fn() -> JsonPath,
        R: FnMut(Error) -> Action,
    {
        Action::Continue
    }

    /// Validate against everything in the specification.
    fn completely<P, R>(&self, root: &Root<'a>, path: P, report: &mut R) -> Action
    where
        P: Fn() -> JsonPath,
        R: FnMut(Error) -> Action,
    {
        self.validate_minimally(root, path, report)
    }
}

alteous:
> Why is it important to (optionally) terminate validation early?

I've decided that, actually, early termination is not important and optimises for the wrong scenario. The most common case will / should be that the glTF is OK, in which case doing any sort of validation means we have to visit every item. Terminating early introduces lots of branches which are not necessary in the most common case and will hence hinder performance.

Thanks for reviewing this! :)

PR #61, opened 2017-07-31
Implement ExactSizeIterator for common iterator types

alteous:
Third time lucky?


alteous:
That, and to be able to check if the iterator is empty. An example would be checking whether a node has children. Returning `Option` in that cases seems unreasonable.

PR #62, opened 2017-07-31
Add prelude module

alteous:
Not sure about this change because it doesn't show up well in the docs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,4 +160,15 @@ pub mod prelude {
     pub use ::texture::Texture;
 }
 
-pub use self::prelude::*;
+pub use self::prelude::Animation;
+pub use self::prelude::Accessor;
+pub use self::prelude::Buffer;
+pub use self::prelude::Camera;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -144,5 +144,20 @@ pub mod texture;
...
+    pub use ::image::Image;
+    pub use ::import::{Data, DynamicImage, Import};
+    pub use ::material::Material;
+    pub use ::mesh::{Mesh, Primitive};
+    pub use ::scene::{Node, Scene};
+    pub use ::skin::Skin;
+    pub use ::texture::Texture;
+}
+
+pub use self::prelude::*;
```
bwasty: For the docs it might help to explicitly list everything here instead of using `*`.

alteous: That's a good idea.

alteous:
Closed since [implemented here](https://github.com/alteous/gltf/blob/master/src/lib.rs#L156-L166).

Issue #57, opened 2017-07-29
Fix Primitive::material default case


bwasty:
Another case where there's an [`Option`](https://docs.rs/gltf/0.7.0/gltf/mesh/struct.Primitive.html#method.material), but the spec defines a default value:
> If material is not specified, then a default material is used.

> The default material, used when a mesh does not specify a material, is defined to be a material with no properties specified. All the default values of material apply. Note that this material does not emit light and will be black unless some lighting is present in the scene.

([src](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes))

Issue #56, opened 2017-07-29
Fix Texture::sampler default case


bwasty:
Analogous to #48, `sampler` should not return an `Option`, but a `Sampler` with default values.
[Docs](https://docs.rs/gltf/0.7.0/gltf/texture/struct.Texture.html#method.sampler)
[Spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#texture): 
> The index of the sampler used by this texture. When undefined, a sampler with repeat wrapping and auto filtering should be used.

alteous:
Thanks for reporting!

Issue #48, opened 2017-07-22
Fix PbrMetallicRoughness default case


alteous:
Currently the `Material` type has `Option<PbrMetallicRoughness>` instead of `PbrMetallicRoughness` with default values. This is an oversight which needs to be fixed.

alteous:
Fixed in the next release.

PR #63, opened 2017-07-31
Wrapper: Return default values instead of Option

alteous:
This commit addresses issues #48, #56, and #57 .


src/material.rs
```diff
@@ -12,30 +12,49 @@ use {extensions, json, texture, Gltf};
...
+            gltf: gltf,
+            index: None,
+            json: &DEFAULT_MATERIAL,
+        }
+    }
+
+    /// Returns the internal JSON index if this `Material` was explicity defined.
+    ///
+    /// This function returns `None` if the `Material` is the default material.
+    pub fn index(&self) -> Option<usize> {
```
alteous: @bwasty Do you think this is a sensible strategy? I considered an enum representation at first but wasn't sure whether it would be actually beneficial:
```rust
enum Material<'a> {
    Default {
        gltf: &'a Gltf,
    },
    Custom {
        gltf: &'a Gltf,
        index: usize,
        json: &'a json::material::Material,
    },
}
```
```

bwasty: I think it's ok. Just requires a tiny change for my `Material` struct (also stores the index for sharing it).

alteous:
Yeah, well spotted... not sure where I got 61 from!

alteous:
Merged manually in order to fix the typo in the commit message.

PR #75, opened 2017-08-12
Add getter for position bounds to Primitive

bwasty:
* min/max was required everywhere when it should only be required for `POSITION`
*  add tests and validation

NOTE: the `[patch.crates-io]` syntax isn't stable yet but will be soon ([2017-08-31](http://doc.crates.io/specifying-dependencies.html#overriding-dependencies)). It might be possible to leave it in permanently - dependent crates don't seem to be affected and would get the right version from crates.io.


bwasty:
At least the nightly build succeeded :)

alteous:
I like the changes in general. The `bounds()` method belongs to a better abstraction IMO - what happens when the bounds are `[f32; 2]`?

Please keep this PR alive for now.

bwasty:
`[f32; 2]` would be invalid I think as `POSITION` has to be `VEC3` and the bounds are supposed to be `per component`. Relevant spec sections:

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#accessors-bounds

alteous:
I meant to say for accessors other than `POSITION`.

bwasty:
Ah, ok. I'm not aware of a use case where you'd want them for other accessors (especially before loading the binary data). This is meant to be a convenience method for the common case of getting bounding boxes (the main reason the fields exist I believe). I think for other cases it would be ok that the user has to parse the data himself.

alteous:
See https://github.com/alteous/gltf/blob/simplify/gltf-utils/src/lib.rs#L298-L312 for a 'better abstraction'.

bwasty:
Can this be used without a `Source`?

alteous:
Nope... Had not considered that.

alteous:
Perhaps just returning `Option<Bounds>` is OK after all.

```rust
impl<'a> Primitive<'a> {
    /// Returns the `(min, max)` values of the `POSITION` attribute, where provided.
    ///
    /// # Notes
    ///
    /// When the `POSITION` attribute is provided, this function will always return `Some`.
    pub fn position_bounds(&self) -> ([f32; 3], [f32; 3]) {
        ...
    }

    /// Returns the `(min, max)` values of the index values, where provided.
    pub fn index_bounds(&self) -> (u32, u32) {
        ...
    }
}

File: gltf-json/src/accessor.rs [hunk 1]
```diff
@@ -190,36 +190,36 @@ pub struct Accessor {
...
     pub extras: Extras,
-    
+
     /// Specifies if the attribute is a scalar, vector, or matrix.
     #[serde(rename = "type")]
     pub type_: Checked<Type>,
 
     /// Minimum value of each component in this attribute.
-    pub min: Vec<f32>,
+    pub min: Option<Vec<f32>>,
```
alteous: This ought to be replaced with `serde_json::Value` or something equivalent.

gltf-json/src/mesh.rs
```diff
@@ -103,36 +103,110 @@ pub struct Mesh {
...
+                    }
+                }
+                else {
+                    report(max_path, Error::Missing);
+                }
+            }
+        }
+
+        // Generated validation
+        fn validate_completely<P, R>(&self, root: &::Root, path: P, report: &mut R)
```
alteous: This can be left out.

gltf-json/src/path.rs
```diff
@@ -10,8 +10,8 @@
 use std::fmt;
 
 /// An immutable JSON source path.
-#[derive(Clone, Debug)]
-pub struct Path(String);
+#[derive(Clone, Debug, PartialEq)]
+pub struct Path(pub String);
```
alteous: I'd rather keep this private for now. Would `Deref<Target = str>` or a function `into_string()` be a reasonable replacement?

src/accessor.rs
```diff
@@ -124,12 +124,12 @@ impl<'a> Accessor<'a> {
     }
 
     /// Returns the minimum value of each component in this attribute.
-    pub fn min(&self) -> &[f32] {
+    pub fn min(&self) -> &Option<Vec<f32>> {
```
alteous: Again, this ought to be `serde_json::Value` or otherwise.

src/mesh.rs
```diff
@@ -371,6 +381,25 @@ impl<'a> Primitive<'a> {
         self.json
     }
 
+    /// Returns the bounds (min/max) of the POSITION attribute if there is one, otherwise `None`.
+    /// May panic for invalid glTF files. Use json::validation::Validate::validate_minimally
+    /// to handle this gracefully.
+    pub fn bounds(&self) -> Option<Bounds> {
```
alteous: I'd prefer to return a tuple `(min, max)` instead of an explicit `Bounds` type. It's up for debate though.

```rust
let (min, max) = primitive.bounds();
```

alteous: Also, for future proofing, I think `position_bounds` would be more appropriate.

src/mesh.rs
```diff
@@ -467,16 +496,17 @@ impl<'a> Loaded<'a, Primitive<'a>> {
...
-            .iter()
-            .find(|&(ref key, _)| key.as_ref().unwrap() == &semantic)
-            .map(|(_, index)| {
-                self.mesh.gltf
-                    .accessors()
-                    .nth(index.value())
-                    .unwrap()
-                    .loaded(self.source)
-            })
+        if let Some(semantic_index) = self.json.attributes.get(&Checked::Valid(semantic)) {
```
alteous: Whatever works. :wink:

tests/test_wrapper.rs
```diff
@@ -0,0 +1,54 @@
+extern crate gltf;
```
alteous: These tests are great, thanks!

tests/test_wrapper.rs
```diff
@@ -0,0 +1,54 @@
...
+    // file with missing min/max values
+    let json = import_json("tests/minimal_accessor_invalid.gltf");
+    let mut errs = vec![];
+    json.validate_minimally(
+        &json,
+        || json::Path::new(),
+        &mut |path, err| errs.push((path(), err)),
+    );
+    assert_eq!(errs,
+        [(Path("meshes[0].primitives[0].attributes[\"POSITION\"].min".into()), Error::Missing),
```
alteous: Is this the reason for the `struct Path(String)` to `struct Path(pub String)` change?

bwasty: Yes.

Cargo.toml
```diff
@@ -21,6 +21,9 @@ travis-ci = { repository = "alteous/gltf" }
 gltf-json = "0.1"
 lazy_static = "0.2"
 
+[patch.crates-io]
```
alteous: Is this necessary any more?

alteous: I think one can specify both `path` and `version` in the manifest format and `cargo` will choose accordingly.

tests/test_wrapper.rs
alteous: If we need to construct `Path`s for testing, we could always just make a `pub(crate) fn from(String)`.

Cargo.toml
bwasty: I didn't know that - I thought `patch` was the new feature to address this, but the docs specifically mention subcrates within one repo under `path`.

src/mesh.rs
bwasty: I had a tuple initially, but changed to struct because you can't see which is `min` and which is `max` from the declaration alone when using a tuple.

Will change the name.

alteous: That actually makes a lot more sense. How about we make `Bounds` generic instead? So, instead of `fn bounds(&self) -> Bounds` we have `fn position_bounds(&self) -> Bounds<[f32; 3]>`?

alteous: [It's no less convenient.](https://play.rust-lang.org/?gist=7a7f0acc9d2b06dd3c221f736899a3a4&version=stable)

bwasty:
Rebasde + addressed most comments, will do the rest later (the tests need some rewriting/moving).

tests/test_wrapper.rs
bwasty: I've tried to make `pub(crate)` work by moving the tests to `gltf-json/tests/test_validation.rs` (more fitting place anyway), but it's still a different crate unfortunately. What's so bad about having the field public? The alternative would be to move the test into `gltf-json/src/mesh.rs` I guess.

bwasty:
I pulled in some changes from my `coverage` branch to get a complete and successful build - please have a look at the last commit (the doc tests should of course be fixed before release, but since they're unrelated I mostly added `ignore` for now).

Cargo.toml
```diff
@@ -17,6 +17,9 @@ include = [
 [badges]
 travis-ci = { repository = "alteous/gltf" }
 
+[workspace]
+members = ["gltf-derive", "gltf-json", "gltf-importer", "gltf-utils"]
```
alteous: :+1:

.travis.yml
```diff
@@ -8,8 +8,8 @@ rust:
 sudo: false
 
 script:
-    - cargo build --verbose --all-features
-    - cargo test --verbose --all-features
+    - cargo build --verbose --all-features --all
```
alteous: :+1:

gltf-json/src/path.rs
```diff
@@ -20,7 +20,7 @@ impl Path {
     ///
     /// Basic usage:
     ///
-    /// ```
+    /// ```rust,ignore
```
alteous: Why were these failing?

tests/test_wrapper.rs
```diff
@@ -1,54 +1,23 @@
...
-    let file = fs::File::open(filename).unwrap();
-    let reader = io::BufReader::new(file);
-    gltf::json::from_reader(reader).unwrap()
-}
 
 #[test]
 fn test_accessor_bounds() {
     // file derived from minimal.gltf with changed min/max values
-    let gltf = gltf::Gltf::from_json(import_json("tests/minimal_accessor_min_max.gltf"));
+    let file = fs::File::open("tests/minimal_accessor_min_max.gltf").unwrap();
```
alteous: `include_bytes!` would save some headache here - just FYI.

gltf-json/src/path.rs
bwasty: (since I'm on mobile:) https://travis-ci.org/bwasty/gltf/jobs/267558880

alteous: Thanks. I'll sort it out later.

PR #76, opened 2017-08-14
Reorganise and simplify

alteous:


gltf-importer/src/lib.rs
```diff
@@ -241,73 +152,58 @@ fn load_external_buffers(
...
+        for (index, buffer) in json.buffers.iter().enumerate() {
+            let path = || {
+                json::Path::new()
+                    .field("buffers")
+                    .index(index)
+                    .field("uri")
+            };
+            match index {
+                0 if has_bin => if buffer.uri.is_some() {
+                    errs.push((path(), Reason::Missing));
```
bwasty: Do I understand this right that validation will fail if a GLB file contains buffer/image data, but also references external data? Why?

src/lib.rs
```diff
@@ -157,42 +160,58 @@ pub use self::animation::Animation;
...
-/// [`Loaded`]: struct.Loaded.html
-pub trait Source: std::fmt::Debug {
-    /// Return the buffer data referenced by the given `Buffer`.
-    ///
-    /// This method must not fail.
-    fn source_buffer(&self, buffer: &Buffer) -> &[u8];
+/// Returns `true` if the data begins with the `b"glTF"` magic string.
+pub fn is_glb(slice: &[u8]) -> bool {
+    slice.starts_with(b"glTF")
 }
```
bwasty: This function is duplicated, but neither one is used by the library.

src/mesh.rs
```diff
@@ -376,195 +185,63 @@ impl<'a> Primitive<'a> {
...
             .iter()
-            .find(|&(ref key, _)| key.as_ref().unwrap() == &semantic)
-            .map(|(_, index)| {
-                self.mesh.gltf
-                    .accessors()
-                    .nth(index.value())
-                    .unwrap()
-                    .loaded(self.source)
-            })
+            .find(|&(key, _)| key.as_ref().unwrap() == semantic)
```
bwasty: Why use `find` on a `HashMap`? `get` would be the "natural" way.

src/skin.rs
```diff
@@ -64,6 +64,21 @@ impl<'a> Skin<'a> {
         &self.json.extras
     }
 
+    /// Returns the accessor containing the visits the 4x4 inverse-bind matrices.
```
bwasty: That sentence sounds a bit strange ("the visits"?)

alteous: Should be 'that' - will fix.

src/mesh.rs
alteous: Using `get` means checking against a `&Checked<Sematic>` rather than a plain `&Semantic`. That's the only reason.

gltf-importer/src/lib.rs
alteous: The glTF JSON slightly different rules for buffers depending on whether it comes from a .gltf file or a .glb file. The former requires URIs for all buffers, whereas the latter requires the first (and only the first) to have no URI if and only if a BIN chunk is provided.

Currently the buffer URI isn't exposed to user although it needs to be. I'm deciding between two ideas: a `Data` enum with `Bin` and `Uri` variants, or returning a special URI (e.g. `"#bin"`) to denote an embedded buffer.

src/lib.rs
alteous: Oops, that's unintentional - will fix.

src/mesh.rs
alteous: I'll replace it with `Checked<Semantic>` instead as it's more clear what's going on.

gltf-importer/src/lib.rs
alteous: The `"#bin"` idea is inspired by these [examples of URI references](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Examples_of_URI_references) - see bottom.

bwasty: An enum sounds better - what could the user do with a special fragment except match against and then ignore it?

alteous: The special fragment would likely be matched and ignored as you say. When loading buffers, the `#bin` fragment means 'the buffer data is in the BIN chunk'.

src/mesh.rs
alteous: The alternative is
```rust
        self.json.attributes
            .get(&json::validation::Checked::Valid(semantic.clone()))
            .map(|index| self.mesh.gltf.accessors().nth(index.value()).unwrap())
```

bwasty: Hmm, so the changed parameter type makes a clone necessary...I wonder with how many attributes the HashMap will be more efficient anyway :)

gltf-importer/src/lib.rs
```diff
@@ -66,143 +68,52 @@ pub enum Error {
...
-            }
-        };
-
-        Ok((header, json_chunk, blob_chunk))
-    }
+/// Imports glTF 2.0
+pub fn import<P>(path: P) -> Result<(Gltf, Buffers), Error>
+    where P: AsRef<Path>
+{
+    import_impl(path.as_ref(), Default::default())
```
bwasty: I just noticed that the `config` is not configurable here. Any reason for that?

gltf-importer/src/lib.rs
```diff
@@ -241,73 +152,58 @@ fn load_external_buffers(
         .collect()
 }
 
-fn validate_standard(json: &gltf::json::Root, config: &Config) -> Result<(), Error> {
+fn validate_standard(
```
bwasty: Make this public perhaps? Same for the `_binary` version.

gltf-importer/src/lib.rs
alteous: It seemed odd to have a `Config` struct with only one parameter, so I left it out to think about it some more. There's been no update since though so I guess we could just add an `import_with_config()` function?

gltf-importer/src/lib.rs
alteous: It's an implementation detail that I'd rather keep private for now. Is there any reason why it should be exported?

bwasty: Not sure if this is the right one to export - I've thought about alternate importers (-> HTTP). What I imagined is one public function to import a .gltf or .glb from binary data, and another one to validate the result (all just for convenience of course, as far as I see it wouldn't be too much code as it is now).

gltf-importer/src/lib.rs
bwasty: 👍

PR #83, opened 2017-08-27
Make sample models submodule non-shallow again

bwasty:
Let's see if the build succeeds...


bwasty:
Not successful, but it got further :)

alteous:
No, that worked! Thanks!

The build failed because there's a problem with the VC sample model.

Issue #93, opened 2017-08-30
Add scene::Transform type


alteous:
Would be returned by a new function `Node::transform`. `Node::matrix/translation/rotation/scale` would be deprecated.

```rust
enum Transform {
    Matrix([[f32; 4]; 4]),
    Decomposed {
        translation: [f32; 3],
        rotation: [f32; 4],
        scale: [f32; 3],
    },
}
```

alteous:
Relevant to #67

Issue #21, opened 2017-05-19
Validate matrices to be decomposable to TRS


bwasty:
From the 2.0 draft spec:
> Matrices must be decomposable to TRS. This implies that transformation matrices cannot skew or shear.

([Source](https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#transformations))

I think it would be good to check the matrices on load. Perhaps this could be optional, so it can be skipped e.g. on release builds.
Relevent discussion on the spec repo: https://github.com/KhronosGroup/glTF/issues/892

alteous:
Good idea. I'm not sure how this could be implemented though.

alteous:
What first comes to mind is transforming the vertices of a simple solid like a tetrahedron and computing the change in the edge lengths. If any scaling takes place then all sides should be scaled by the same amount or else there would be some shear or skew taking place.

bwasty:
There are formulas to extract rotation/translation/scale/skew/shear from matrices:
https://math.stackexchange.com/questions/237369/given-this-transformation-matrix-how-do-i-decompose-it-into-translation-rotati
http://stackoverflow.com/questions/5107134/find-the-rotation-and-skew-of-a-matrix-transformation
http://stackoverflow.com/questions/15420693/how-to-get-rotation-translation-shear-from-a-3x3-homography-matrix-in-c-sharp

My first idea was to extract T/R/S, then re-multiply and assert they're the same. But I'm not sure that's enough to be certain. Second idea: extract only skew and shear and assert that they are not present. Also not entirely sure that's enough.

alteous:
I'm not sure if there's been any consensus or further discussion on this issue. The specification doesn't seem to put any restrictions on the node transform matrices.

bwasty:
I still see it in the spec, just slightly reworded:

> When `matrix` is defined, it must be decomposable to TRS. This implies that transformation matrices cannot skew or shear.

([src](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#transformations))

alteous:
I think the validation may be as simple as checking the bottom row for non-zero values in the matrix. For example this matrix would be acceptable:

```
[ s r r t ][ x ]
[ r s r t ][ y ]
[ r r s t ][ z ]
[ 0 0 0 w ][ 1 ]
```

But this matrix would not:

```
[ s r r t ][ x ]
[ r s r t ][ y ]
[ r r s t ][ z ]
[...]

PR #94, opened 2017-08-31
Add scene::Transform

alteous:
This addresses issue #93

### DIFF
--- a/src/scene.rs
+++ b/src/scene.rs
@@ -24,7 +24,10 @@ const IDENTITY: [f32; 16] = {
 #[derive(Clone, Debug)]
 pub enum Transform {
     /// 4x4 transformation matrix in column-major order.
-    Matrix([[f32; 4]; 4]),
+    Matrix {
+        /// 4x4 matrix.
+        matrix: [[f32; 4]; 4],
+    },
 
     /// Decomposed TRS properties.
     Decomposed {
--- a/src/scene.rs
+++ b/src/scene.rs
@@ -8,11 +8,38 @@
 // except according to those terms.
 
 use json;
-use std::slice;
+use std::{mem, slice};
 
 use {Camera, Gltf, Mesh, Skin};
 
-///  A node in the node hierarchy. When the node contains `skin`, all
+/// 4x4 identity matrix.
+const IDENTITY: [f32; 16] = {
+    [1.0, 0.0, 0.0, 0.0, 
+     0.0, 1.0, 0.0, 0.0, 
+     0.0, 0.0, 1.0, 0.0, 
+     0.0, 0.0, 0.0, 1.0]
+};
+
+/// The transform for a `Node`.
+#[derive(Clone, Debug)]
+pub enum Transform {
+    /// 4x4 transformation matrix in column-major order.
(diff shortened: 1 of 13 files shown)


alteous:
We really need #21 before this is published.

src/scene.rs
```diff
@@ -7,12 +7,108 @@
...
+                (t * r * s).into()
+            },
+        }
+    }
+
+    /// Returns the decomposed representation of this transform.
+    ///
+    /// If the transform is `Matrix`, then the decomposition is extracted from the
+    /// matrix.
+    pub fn decomposed(self) -> ([f32; 3], [f32; 4], [f32; 3]) {
```
bwasty: This needs good tests.

src/scene.rs
```diff
@@ -127,20 +228,40 @@ impl<'a> Node<'a> {
...
     pub fn translation(&self) -> [f32; 3] {
         self.json.translation
     }
 
+    /// Returns the node's transform.
+    pub fn transform(&self) -> Transform {
+        if let Some(matrix) = self.json.matrix.clone() {
+            unsafe {
+                Transform::Matrix {
+                    matrix: mem::transmute(matrix),
```
bwasty: This loos like it does two extra copies compared to the deprecated `matrix` method (clone + transmute) - is that right?

src/scene.rs
```diff
@@ -7,12 +7,108 @@
...
+    /// Returns the matrix representation of this transform.
+    ///
+    /// If the transform is `Decomposed`, then the matrix is generated with the
+    /// equation `matrix = translation * rotation * scale`.
+    pub fn matrix(self) -> [[f32; 4]; 4] {
+        match self {
+            Transform::Matrix { matrix } => matrix,
+            Transform::Decomposed { translation, rotation, scale } => {
+                let t = Matrix4::from_translation(translation.into());
+                let r = Matrix4::from(Quaternion::from(rotation));
```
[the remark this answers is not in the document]
alteous: Well spotted!

src/scene.rs
alteous: I don't think so.

`mem::transmute` moves bytes and `matrix` is `Copy` anyway.

alteous: Oops... I pressed review button by accident. Not sure what that does.

src/scene.rs
alteous: Time to write some tests.

alteous: At the moment I'm testing by loading the sample models with `three`. All seems to be in order.

What kind of tests should be written? Sometimes the quaternion doesn't match the original but is an equivalent rotation.

alteous:
Latest commit is unrelated but important.

src/scene.rs
bwasty: Hmm...for this kind of math I'd try to find a few examples with as many edge cases as possible (not exactly sure what this would be here), possible compute the expected result with an alternative implementation and use some "assert almost equals" function (for numeric issues).
Not sure about the quaternion - re-multiplying maybe?

bwasty: Or just dump the values from some tests with `three` into tests :)

Issue #92, opened 2017-08-30
Clean up documentation


alteous:
Some formatting, such as [`Material::alpha_mode`](https://docs.rs/gltf/0.9.0/gltf/material/struct.Material.html#method.alpha_mode) has been lost from the [JSON equivalent](https://docs.rs/gltf-json/0.9.0/gltf_json/material/struct.Material.html).

Issue #91, opened 2017-08-30
Morph targets is missing fields


alteous:
Blatantly obvious in the docs. https://docs.rs/gltf/0.9.0/gltf/mesh/struct.MorphTargets.html

Issue #90, opened 2017-08-30
Note material default values in documentation


alteous:
#### Incomplete table of undocumented items
| Item  | Default  |
|---|---|
| `PbrMetallicRoughness::base_color_factor`  | `[1.0, 1.0, 1.0, 1.0]` |
| `PbrMetallicRoughness::metallic_factor`  |  `1.0`  |
| `PbrMetallicRoughness::roughness_factor` | `1.0`  |
| `Material::emissive_factor` | `[0.0, 0.0, 0.0]` |

Issue #89, opened 2017-08-30
Add Gltf::from_glb function


alteous:
Or something equivalent. The connection between `Glb` and `Gltf` is not obvious.

PR #96, opened 2017-09-05
V0.9.1

alteous:
Closes issues #93, #92, #91, #90, #89, #88, #87, #86, #85, and #84.


src/lib.rs
```diff
@@ -60,49 +60,44 @@
 //! ```
 
 extern crate cgmath;
+extern crate gltf_json as json;
 #[macro_use]
 extern crate lazy_static;
 
-/// Contains (de)serializable data structures that match the glTF JSON text.
-pub extern crate gltf_json as json;
```
alteous: Technically a breaking change. Should this be 0.10.0 instead?

alteous: Actually, I'll put this back as I'd rather have a 0.9.1 version than a 0.10 version. Let's deprecate instead.

alteous:
This is ready for review now.

alteous:
The Travis build passed! :open_mouth:

PR #100, opened 2017-09-08
Fix Transform::decomposed

### LINKED ISSUE #99: Transform::decomposed is wrong for matrices with negative determinant
Eg. for `Transform::Matrix { matrix: Matrix4::from_nonuniform_scale(1.0, 1.0, -1.0) })`, it currently produces a rotation of [0, 0, 0, √2] and a scale of [1, 1, 1].

One way to fix it is to negate one of the columns and the corresponding scaling factor, pushing the reflection from the rotation part to the scaling part. Something like

    if m.determinant() < 0 {
        m[0] = -m[0];
        sx = -sx;
    }

Note that there are at least four choices for how to do this: negate the first column, negate the second column, negate the third column, negate all the columns. Because of this, and the fact that the map from quaternions to matrices is a double cover, I think the docs for `decomposed` should say "Returns _a_ decomposed representation...".

alteous:
Fixes #99

### DIFF
--- a/src/scene.rs
+++ b/src/scene.rs
@@ -347,7 +347,11 @@ mod tests {
         let matrix = Transform::Decomposed { translation, rotation, scale }.matrix();
         let (translation, rotation, scale) = Transform::Matrix { matrix }.decomposed();
         let check = Transform::Decomposed { translation, rotation, scale }.matrix();
-        assert_relative_eq!(Matrix4::from(check), Matrix4::from(matrix));
+        assert_relative_eq!(
+            Matrix4::from(check),
+            Matrix4::from(matrix),
+            epsilon = 0.05
+        );
     }
 
     fn test_decompose_rotation(rotation: [f32; 4]) {
--- a/src/scene.rs
+++ b/src/scene.rs
@@ -17,7 +17,6 @@ use {Camera, Gltf, Mesh, Skin};
 type Matrix3 = cgmath::Matrix3<f32>;
 type Matrix4 = cgmath::Matrix4<f32>;
 type Quaternion = cgmath::Quaternion<f32>;
-type Vector3 = cgmath::Vector3<f32>;
 
 /// 4x4 identity matrix.
 const IDENTITY: [f32; 16] = {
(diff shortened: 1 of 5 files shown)


alteous:
@scurest: Would you mind reviewing this for us?

src/scene.rs
```diff
@@ -344,3 +347,51 @@ impl<'a> Iterator for Children<'a> {
...
+
+        let (t, r, s) = Transform::Matrix { matrix }.decomposed();
+        check(translation, rotation, scale, t, r, s);
+    }
+
+    #[test]
+    fn decompose_matrix_with_negative_determinant() {
+        let translation = [1.0, 2.0, 3.0];
+        let rotation = [0.0, 0.0, 0.0, 1.0];
+        let scale = [1.0, 1.0, -1.0];
```
alteous: When the x or y scale is negative the decomposition is incorrect.

src/scene.rs
```diff
@@ -79,14 +79,17 @@ impl Transform {
                     mem::replace(&mut m[3][2], 0.0),
                 ];
                 let sx = Vector3::new(m[0][0], m[0][1], m[0][2]).magnitude();
+                let sy = Vector3::new(m[1][0], m[1][1], m[1][2]).magnitude();
+                let mut sz = Vector3::new(m[2][0], m[2][1], m[2][2]).magnitude();
+                if cgmath::Matrix4::from(m).determinant() < 0.0 {
+                    sz = -sz;
```
alteous: Is any further negation required here?

scurest: No, that looks good. If you want, you can cut `m` down to a `Matrix3` right after you extract the translation though, instead of when you form `r`, which might make taking the determinant a little faster and the code a little nicer (eg. `let sx = m.x.magnitude()`, `m.x /= sx`).

src/scene.rs
scurest: Yeah, you can't check for equality of TRS components  because a TRS decomposition isn't unique; you have a choice of where to put the sign on the scale, you have a choice between q and -q for the rotation, etc. You could put `t`, `r`, and `s` into a new `Transform::Decomposed`, convert it to a matrix, and check for approximate equality with `matrix` instead.

alteous: Round trip - I like that idea.

I wonder it's worth always picking the quaternion whose scalar value is positive, just so that decomposition is deterministic.

According to the specification, a negative z scale means the geometry must be rendered with the winding order reversed. It doesn't make too much sense to have a negative x or y scale as well, so perhaps we should ignore these cases and just make a note in the docs such as "matrices must not scale by a negative factor in the x or y axes".

src/scene.rs
alteous: Nice, will do.

scurest: Using `signum` is very pretty!

src/scene.rs
scurest: If you specify that the x and y scales are always positive and fix the quaternion you choose I think the decomposition is fully determined. For the quaternion you need to do slightly more than making the scalar positive because of the case when the scalar is zero. Something like  "the first component of the quaternion, counting from the end, that is non-zero is positive".

The new `test_decompose` looks good 👍 I think it'd be good to test a scale where the negative value is on the x or y component, since having it on the z component is kind of the "happy path" for `decomposed` and exercising the non-happy path would give me more confidence in the test.

alteous:
I've written a few more tests, some of which fail. See the [Travis build](https://travis-ci.org/gltf-rs/gltf/jobs/273810219#L570) for the details. Do these test failures suggest any cause for concern?

scurest:
Not really. I'm pretty sure they're all due to rounding error in the floating point calculations. The entry where the matrices differ in each test should be zero and it is notoriously hard to compare numbers around zero. If you vary the angle and direction of rotation or the `max_relative` and `epsilon` arguments for the `assert_relative_eq` the number of tests that fail varies a lot.

I think what I'd do is use a very generous `epsilon` in the relative equality test (`epsilon = 0.05` works). I know that sounds fairly coarse, but even with it, your new tests would have caught #99.

alteous:
Excellent. Thank you for your help! :smile:

PR #105, opened 2017-09-13
gltf-editor (work in progress)

alteous:
A simple tool for querying data structures in a .gltf file. Useful for when .gltf is very large and / or minified.

There is a useable text interface (`cargo run --bin gltf-editor -- path/to/asset.gltf --tui`) and a planned graphical interface using [`GtkTreeView`](https://developer.gnome.org/gtk3/stable/GtkTreeView.html) or otherwise. Eventually, it would be nice to be able to edit values using either interface.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,10 @@ default = []
 extras = ["gltf-json/extras"]
 names = ["gltf-json/names"]
 
+[[bin]]
+name = "gltf-editor"
+path = "bin/gltf-editor.rs"
+
 [[example]]
 name = "gltf-display"
 path = "examples/display/main.rs"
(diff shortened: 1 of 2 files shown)


alteous:
This might be best located outside of the crate alongside `gltf-viewer`.

bwasty:
I've thought about a tree-view based editor before, but https://github.com/AnalyticalGraphicsInc/gltf-vscode is pretty awesome already, so I'm not sure it's worth the effort.
About the TUI: I stumbled on https://github.com/kkawakam/rustyline yesterday - might be helpful?

alteous:
I'll check those out, thanks!

Issue #38, opened 2017-06-26
Travis times out on cloning sample models on Mac build


bwasty:
Had two failed builds in my project (got a dependecy on the `wrapper` branch) and saw it also happened once here: 
https://travis-ci.org/alteous/gltf/jobs/242077737

Perhaps it would be an alternative to download the zip file (https://github.com/KhronosGroup/glTF-Sample-Models/archive/master.zip). The repo is a 1GB after cloning, the .zip only 540MB and downloads in ~2min for me.

alteous:
That would be a good alternative. This could be merged with https://github.com/alteous/gltf/issues/39.

alteous:
N.B. the download link is actually https://codeload.github.com/KhronosGroup/glTF-Sample-Models/zip/master.

alteous:
Since glTF-Sample-Models was made a 'shallow' submodule (i.e. only the last commits are cloned) this issue seems to have stopping occuring. I hope this will continue to be the case in the future.

At the moment I'm not experienced enough with version control systems to figure out a better solution. Downloading the zipped models worked on my own OS (Kubuntu) but there seems to be no cross-platform way without depending on significantly more crates.

Has the download time improved for you since this change was made?

bwasty:
Actually I had timeouts again, though I'm not sure I was on the correct branch (didn't find that commit in the history)...let me try again.

bwasty:
Hmm, no luck: https://travis-ci.org/bwasty/gltf-viewer/jobs/252109361
It's pretty weird, considering that on Linux the whole build takes under 4 minutes.
I don't have a good idea at the moment (except commenting out the OS X build while using gltf from git). There is https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received, but this requires some more changes to the travis config I guess (since there is no explicit command one could put `travis_wait` before).

bwasty:
Idea I had while once again waiting for the samples submodule (...happens locally for each `cargo update`):
extract the samples to a [dev dependency](http://doc.crates.io/specifying-dependencies.html#development-dependencies) somehow. Not sure about the details though.

alteous:
> extract the samples to a dev dependency somehow

How would this help?

alteous:
What would be really great is if we could package up pre-compiled crates. I have a slightly crazy idea of using `include_bytes!` to bake the models into a library. That way we only package the models we need and without any git metadata.

bwasty:
[...]

PR #107, opened 2017-09-18
Add travis_wait to travis config

bwasty:
might solve #38


alteous:
> Ran for 4 hrs 51 min 2 sec 

But at least it passed!

Issue #106, opened 2017-09-15
Buffer::uri return type


alteous:
In rare scenarios, it's possible that our strategy to return `"#bin"` (meaning the `BIN` section from binary glTF) from `Buffer::uri` might conflict. See https://github.com/KhronosGroup/glTF/pull/1100#issuecomment-329709830 for the details.

I wanted to avoid an enum like `Image::Data` because it's a special but predictable scenario. Opinions?

alteous:
```rust
enum Location<'a> {
    Bin,
    Uri(&'a str),
}
```

Issue #12, opened 2017-03-24
Plans for 1.0.0


alteous:
Now that the implementation of the glTF 1.0 specification is complete, the next steps are to consolidate the library to a production standard. `1.0.0` will be the version where this condition is met.

Below is an non-exhaustive list of things I would expect from a `1.0.0` release:

- [x] Guaranteed backward compatibility for all `1.x.x` releases
- [x] Comprehensive documentation
- [x] glTF 2.0 support
- [x] `KHR_binary_glTF` support
- [x] `Validate` implementations for all 2.0 data structures
- [x] Basic wrapper library for the 2.0 implementation
- [x] ~~Conversion to and from 1.0 / 2.0 assets~~ (see #14)

fst3a:
In my fork I have altered the API a little, changed few bits so that they take generic Read objects instead of Path so that it's possible to load models from memory and such.  Additionally, I don't think that slice calls in v2 are needed.

alteous:
N.B. in the [upcoming wrapper interface](https://github.com/alteous/gltf/blob/wrapper/src/import.rs#L108) we're planning to take generic `Read`.

fst3a:
Oh also I don't think there's a reason to load whole resource in memory and then call from_slice on it.  You could do something like this https://github.com/fst3a/gltf/blob/master/src/v1/gltf.rs#L263

alteous:
> Actual loading of data (images, bins and such) and paths should, in my opinion, be delegated to the user

I mostly agree. The idea behind the `Source` trait is that the user is able to implement their own loader that is called by the importer. This means the importer can make safety guarantees about external data whilst providing a convenient wrapper around the JSON data structures.

Code bloat is a fair concern. Perhaps the library could make the wrapper an opt-out feature.

If all the user wants is to obtain the raw JSON data and validate then that's fine, the appropriate data structures are available in the `json` module.

I like the code snippet, I haven't seen `chain` before. That's a nice trick.

fst3a:
I see.

alteous:
I forgot to mention: The library is moving toward so called 'zero-cost deserialization' of the JSON data structures, which requires `from_slice` or otherwise to obtain the lifetime of the data. The latest code in the `wrapper` branch demonstrates this.

alteous:
[...]

PR #108, opened 2017-09-22
V1.0

### LINKED ISSUE #13: Library structure
This thread exists as a place to discuss the modules, functions, data structures, and their names included in the library. In version `1.0.0` names must be agreed upon and remain immutable for backward compatibility. Until then, we should experiment as much as possible and find out what works best.

#### gltf

 - The main crate, intended for navigating glTF JSON.
 - Intentionally limited in scope.

#### gltf-json

 - Child crate for `gltf`.
 - Exists primarily to reduce compile times for `gltf`.
 - `semver` matches that of `gltf` for convenience but is otherwise unstable.

 #### gltf-importer

- Reference importer with a simple API.
- May be used by run times to read from the file system.
- Handles base 64 decoding and distinguishing binary glTF on behalf of the user.
- Optionally loads buffers and images (to do).

#### gltf-utils

- Utility functions to complement the `gltf` crate.
- Addresses cases where access to pre-existing buffer / image data is required.
- ~~May remain unstable post `gltf 1.0`.~~

alteous:
Closes #13 and #12

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -13,6 +13,15 @@ use Gltf;
 
 pub use json::buffer::Target;
 
+/// The data referenced by a `Buffer`.
+pub enum Data<'a> {
+    /// Buffer data is contained in the `BIN` section of binary `glTF`.
+    Bin,
+
+    /// Buffer data is contained in an external data source.
+    Uri(&'a str),
+}
+
 /// A buffer points to binary data representing geometry, animations, or skins.
 #[derive(Clone, Debug)]
 pub struct Buffer<'a> {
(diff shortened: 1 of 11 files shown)


alteous:
> Ooh - time for 1.0 already?

I think so. I've made this PR mainly for visibility. I don't expect this to be merged within the next week.

alteous:
>  I'd wish the library would make image loading a bit simpler or at least provide a complete example).

How could we address this?  Would a `read_image` function in `gltf-utils` plus `gltf-importer` returning `(Gltf, Buffers, Images)` be suitable?

bwasty:
> >  I'd wish the library would make image loading a bit simpler or at least provide a complete example).
> 
> How could we address this?  Would a `read_image` function in `gltf-utils` plus `gltf-importer` returning `(Gltf, Buffers, Images)` be suitable?

I've also thought about `gltf-utils`, not sure if it has to be in `gltf-importer` as well. Essentially it should encapsulate [this block](https://github.com/three-rs/three/blob/139c22c757d94b75dd8deda90f2934a08b0ed2b3/src/factory/load_gltf.rs#L23-L54) minus the image decoding - give it a `gltf::image::Image`, get raw binary data without caring whether it comes from a file, GLB or Data URI. Not sure in how far a really simple interface is feasible. It doesn't have to be super convenient, but saving most of those lines would be nice.

src/buffer.rs
```diff
@@ -13,6 +13,15 @@ use Gltf;
 
 pub use json::buffer::Target;
 
+/// The data referenced by a `Buffer`.
+pub enum Data<'a> {
```
alteous: Closes #106

alteous:
This is a long way off.

PR #117, opened 2017-10-01
Fix clippy issues

bwasty:
Tip: append [`?w=1`](https://github.com/gltf-rs/gltf/pull/117/files?w=1) to ignore whitespace changes.

I wasn't sure about the `cast_lossless` - theses are the concrete cases: https://gist.github.com/bwasty/21ae877413e4770e4eb5d1b05ebd6284#file-gltf-clippy-txt-L408-L474

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,6 +8,7 @@
 // except according to those terms.
 
 #![deny(missing_docs)]
+#![allow(unknown_lints)]
 
 //! glTF 2.0 loader
 //!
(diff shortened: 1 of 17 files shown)


src/lib.rs
```diff
@@ -8,6 +8,7 @@
 // except according to those terms.
 
 #![deny(missing_docs)]
+#![allow(unknown_lints)]
```
alteous: What does this do?

bwasty: The Rust compiler doesn't know clippy's lints and otherwise shows a warning for the `allows` in other places.

PR #123, opened 2017-10-11
Fix Glb loading on big-endian

### LINKED ISSUE #120: Glb doesn't load on Big-endian system
Tested on router, apparently it is due transmutes in glb loading code.  Currently working on fix (adds dependency on `byteorder` crate).  I think this is worth considering (aswell as getting rid of as many unsafes as required) before 1.0 release.

fst3a:
Fixes #120


alteous:
Does #125 replace this PR? (Can I close this in favour of #125 instead?)

fst3a:
Yes, that is the point of #125 .  I left this open in case You had some strong objections against #125 .

Issue #121, opened 2017-10-11
AccessorIter is unsound can lead to undefined-behavior in safe code


fst3a:
It also will probably not work properly on a BE system, I didn't test, though.

```rust
extern crate gltf;
extern crate gltf_utils;

use gltf_utils::AccessorIter;

use std::{fs, io};

#[derive(Debug)]
pub struct Dummy;

impl gltf_utils::Source for Dummy {
    fn source_buffer(&self, _: &gltf::Buffer) -> &[u8] {
        &[0xff; 1024]
    }
}

fn main() {
    let file = fs::File::open("../tests/minimal_accessor_min_max.gltf").unwrap();
    let mut reader = io::BufReader::new(file);
    let gltf = gltf::Gltf::from_reader(&mut reader).expect("yes")
        .validate_minimally().expect("yes 2");
    let dummy = Dummy;
    let a2 = gltf.accessors().nth(0).unwrap();
    let a12 = gltf.accessors().nth(1).unwrap();

    // With bools
    {
        // UB because a bool that isn't 0 or 1 is invalid in Rust.
        let mut whoops = AccessorIter::<[bool; 2]>::new(a2, &dummy);
        let bools = whoops.next().unwrap();
        println!("{:?}", bools);
        println!("{:?} {:?}",
                 unsafe { *(&bools as *const _ as *const u8) },
                 unsafe { *(&true as *const _ as *const u8) });
    }

    // With refs
    {
        // Twice UB: gets a dummy reference and struct is not repr(C).
        #[derive(Debug, Copy, Clone)]
        #[repr(packed)]
        struct Ref {
            a: &'static u32,
            b: u32,
        }
        let mut whoops = AccessorIter::<Ref>::new(a12, &dummy);
        let refs = whoops.next().unwrap();
        // Oops!
        println!("{:?}", refs);
    }
}
```

fst3a:
Also, in gltf-importer validation strategy can be selected to be skipped which again "safely" does an "unsafe" action (calls unsafe `skip_validation` method).  But I am not sure if this was intended.  If library can function safely even if validation is skipped, perhaps unsafe should be removed from it?

alteous:
>  If library can function safely even if validation is skipped, perhaps unsafe should be removed from the method?

Maybe. The idea is to discourage users from using `skip_validation`, that's why `unsafe` is required! There's nothing *really* unsafe about skipping validation, but the library will panic if the glTF is indeed invalid.

[...]

PR #125, opened 2017-10-13
Completely rework GLB loader and add from_reader method

fst3a:
This PR supersedes #123 but is somewhat more aggressive.  It completely reworks GLB loader to make code more clear to understand and also implements a `from_reader` method in addition to existing `from_slice`, which takes responsibility of loading GLB from any `io::Read` stream for extra user convenience.


src/glb.rs
```diff
@@ -6,47 +6,118 @@
...
 // except according to those terms.
+use byteorder::{LE, ReadBytesExt};
 
 use Error;
 
+use std::{fmt, io};
+
+/// Represents a Glb loader error.
+#[derive(Debug)]
+pub enum GlbError {
```
alteous: I have an aversion toward repeated prefixes and suffixes. `Error` is just fine, as the crate will use `glb::Error` and the user `gltf::glb::Error`.

src/glb.rs
```diff
@@ -6,47 +6,118 @@
...
+    /// Io error occured.
+    IoError(::std::io::Error),
```
alteous: Same here. We know it's an error, so `Io` is enough.

alteous: I like the breakdown of the individual problems.

src/glb.rs
```diff
@@ -6,47 +6,118 @@
...
     pub version: u32,
-
     /// Must match the length of the parent .glb file.
     pub length: u32,
 }
 
-/// The header of the JSON or BIN section of a .glb file.
+/// GLB chunk type.
+#[derive(Copy, Clone, Debug)]
+pub enum ChunkType {
```
alteous: :+1:

src/glb.rs
```diff
@@ -6,47 +6,118 @@
...
+    Json,
+    /// BIN\0 chunk.
+    Bin,
+}
+
+/// Chunk header with no data read yet.
 #[derive(Copy, Clone, Debug)]
-#[repr(C)]
-struct ChunkInfo {
+struct ChunkHeader {
```
alteous: This is better, thanks.

src/glb.rs
```diff
@@ -6,47 +6,118 @@
...
+                length: reader.read_u32::<LE>().map_err(IoError)?,
+            })
+        } else {
+            Err(GlbError::Magic(magic))
+        }
+    }
 
-    /// Either `b"JSON"` or `b"BIN\0"`.
-    kind: [u8; 4],
+    fn size_of() -> usize { 12 }
```
alteous: :+1:

src/glb.rs
```diff
@@ -55,71 +126,134 @@ impl<'a> Glb<'a> {
...
-            }
-            if (*x).length as usize != data.len() {
-                return err("length does not match length of data");
-            }
-            if (*x).version != 2 {
-                return err("Not GLB version 2.0");
+    /// Does the loading job for you.  Provided buf will be cleared before new
+    /// data will be written.  When error happens, if only header was read, buf
+    /// will not be mutated, otherwise, buf will be empty.
+    pub fn from_reader<R: io::Read>(mut reader: R,
```
alteous: Good idea. :+1:

src/glb.rs
```diff
@@ -55,71 +126,134 @@ impl<'a> Glb<'a> {
...
+                // SAFETY: We are doing unsafe operation on a user-supplied
+                // container!  Make sure not to expose user to uninitialized
+                // data if an error happens during reading.
+                //
+                // It is guaranteed by reserve's implementation that the reserve
+                // call will make buf's capacity _at least_ buf_len.
+                //
+                // We do not read contents of the Vec unless it is fully
+                // initialized.
+                unsafe { buf.set_len(glb_len as usize) };
```
alteous: The safety note is a bit overkill but I appreciate the sentiment.

src/glb.rs
```diff
@@ -55,71 +126,134 @@ impl<'a> Glb<'a> {
...
+            .and_then(|json_h| if json_h.length as usize <= data.len() {
+                Ok(json_h)
+            } else {
+                Err(GlbError::ChunkLength {
+                    ty: json_h.ty,
+                    length: json_h.length,
+                    length_read: data.len(),
+                })
+            })
+            // PANIC: We have verified that json_h.length is no greater than
```
alteous: I'm a bit worried to be reading `PANIC` in big letters in the source code. Does this actually `panic!`?

src/glb.rs
```diff
@@ -55,71 +126,134 @@ impl<'a> Glb<'a> {
...
+             GlbError::Version(_) => "unsupported version",
+             GlbError::Magic(_) => "not glTF magic",
+             GlbError::Length { .. } => "could not completely read the object",
+             GlbError::ChunkLength { ty, .. } => match ty {
+                 ChunkType::Json => "JSON chunk length exceeds that of slice",
+                 ChunkType::Bin => "BIN\\0 chunk length exceeds that of slice",
+             },
+             GlbError::ChunkType(ty) => match ty {
+                 ChunkType::Json => "was not expecting JSON chunk",
+                 ChunkType::Bin => "was not expecting BIN\\0 chunk",
```
alteous: The `\\0` part is irrelevant to the user-facing error but it's no big deal.

src/lib.rs
```diff
@@ -125,10 +126,8 @@ pub use self::texture::Texture;
 pub enum Error {
     /// JSON deserialization error.
     Deserialize(json::Error),
-
     /// GLB parsing error.
-    Glb(String),
-
+    Glb(self::glb::GlbError),
```
alteous: Would prefer `Glb(glb::Error)` as discussed at the top.

src/glb.rs
fst3a: No, it explains why it doesn't.

src/glb.rs
fst3a: I understand, but this is made to not to confuse the user, so that user could easily look up BIN\0 chunk type in e.g. gltf specs.

fst3a:
I will try to come up with a fix for AccessorIter, but AccessorIter is a part of separate crate (gltf-utils) anyway.

src/glb.rs
alteous:
```rust
/// FIRE MISSILES: This function does not fire any missiles.
fn foo() {}
```

fst3a:
This satisfies most requested changes except sound AccessorIter (part of gltf-utils).

alteous:
Did you see my PR?

alteous:
([This one](https://github.com/fst3a/gltf/pull/1))

fst3a:
Whoops, no, I didn't see it.

fst3a:
Don't really know how to test AccessorIter so implementation might be incorrect.

fst3a:
Would be nice for somebody else to double-check the implementation against spec and/or express their concerns.  Don't really know how to test this for edge cases.

alteous:
Will double-check with `three` and `gltf-viewer`.

fst3a:
I'm certainly sure we never use signed integers in gltf-utils, even though they appear in the spec, though.  Other than that, we are apparently missing some matrix types?

type | Number of components
-- | --
"SCALAR" | 1
"VEC2" | 2
"VEC3" | 3
"VEC4" | 4
"MAT2" | 4
"MAT3" | 9
"MAT4" | 16

Should AccessorItem be implemented for those as well?

fst3a:
It doesn't appear that we use any signed types anyway, even though they appear in spec.  As for spec it only requires us to have i8 and i16 but not i32 so I didn't implement it.

As for matrices, we're already covered by array implementations.  [[T; 4]; 4] would work because [T; 4] has AccessorItem implemented for it so no need to explicitly implement it for the nested type.

alteous:
Awesome work! Thank you. :smile:

fst3a:
This fixes #120 and #121 .

PR #134, opened 2017-10-27
PrimitiveIterator::joints_u16() retrieves Semantic::Joints(set) instead of Semantic::Colors(set)

Elzair:
I just fixed a quick typo.


alteous:
Thanks for the PR. This has been fixed in the next major release but that's unlikely to be published soon.

I'll merge and publish a new minor release with the patch.

Elzair:
Awesome! Thanks!

PR #135, opened 2017-10-29
AccessorIter fast paths

fst3a:
This PR makes certain operations on `AccessorIter` O(1) and should reduce redundant range checking in `AccessorItem::from_slice`.


alteous:
Nice work, thanks!

PR #136, opened 2017-10-31
Tests for AccessorIter and AccessorItem

fst3a:
This PR adds some tests which test `AccessorIter` and its fast paths.  In addition, this also removes redundant `accessor` field of `AccessorIter` and fixes bug in `AccessorIter::size_hint` which also affects `AccessorIter::count`.


fst3a:
How should AccessorIter's fast paths be forwarded in wrapping types (`Positions`, `Normals` etc)?

alteous:
Great, thanks!

PR #137, opened 2017-10-31
Glb::from_reader revamp

fst3a:
This PR changes refs to Cows to make it possible for Glb to own the data.  The implementation might not be as fast and allocation-low as the old one but this one is shorter and nicer to use.

It also fixes _potential_ oversight of mine in old code where safe code implementing `io::Read::read_exact` could be exposed to uninitialized data which _potentially_ could lead to bad things like leaks of sensitive information, _in theory_.


alteous:
:shipit:

PR #139, opened 2017-11-01
Relax lifetimes in gltf-utils

fst3a:
This PR makes it possible to do things like:
```rust
let primitives: Vec<_> = mesh.primitives().flat_map(|x| x.positions(&bins)).collect();
```
which would currently be an error because of overly strict lifetime requirements.  This is solved by introducing additional lifetime `'s`, which prevents `'a` from being coerced to shortest common of `PrimitiveIterator` and `Source`.


alteous:
Seems reasonable, thanks.

PR #142, opened 2017-11-09
Refactored gltf-util ChannelIterators based on Issue #141

Elzair:
By request in #141  `ChannelIterators` now has a method `outputs` that returns an `enum` containing the iterator for the channel data based on the type of data.

PR #143, opened 2017-11-09
Rework casting iterators and let user obtain the underlying AccessorIters 

fst3a:
This PR removes most old newtype iterators replacing them with type aliases and also adds more casts (these come at cost of more copy paste).  It conflicts with https://github.com/gltf-rs/gltf/pull/142.

There are some things left:

* Should methods try to fail with None as much as they can instead of blowing up the thread at => unreachable!() arms, if user chooses to opt out of validation?


vitvakatu:
I must say that `#[inline(always)]` isn't best thing to do. It force compiler to inline functions even if it not help with optimization. Use `#[inline]` instead - it would give compiler much more flexibility. See [this thread](https://internals.rust-lang.org/t/when-should-i-use-inline/598/) for some info about inlining.
cc @alteous

fst3a:
Compiler inlined them anyway because they're crate-local but I added annotations just in case.

fst3a:
Apparently PrimitiveIterators::indices method is obscured by Primitive::indices, which I didn't notice (sorry).  What should it be renamed to?

alteous:
FYI I'm very busy at the moment. I'll be active again in a couple of days.

gltf-utils/src/lib.rs
```diff
@@ -49,7 +32,7 @@ pub trait Source: fmt::Debug {
 }
 
 /// Extra methods for working with `gltf::Primitive`.
-pub trait PrimitiveIterators<'a> {
+pub trait PrimitiveIterators<'a>: private::Sealed {
```
alteous: What does this do exactly? I imagine it prevents further implementations but I'm not sure how.

gltf-utils/src/lib.rs
```diff
@@ -1,41 +1,24 @@
...
-/// ```
-///
-/// Denormalize an array of integers.
-///
-/// ```rust
-/// use gltf_utils::Denormalize;
-/// let rgb: [u8; 3] = [0, 120, 255];
-/// assert_eq!([0.0, 120.0 / 255.0, 1.0], rgb.denormalize());
-/// ```
-pub trait Denormalize {
```
alteous: I think this should stay for now.

gltf-utils/src/lib.rs
fst3a: By not forwarding Sealed trait from mod "private" to public.  It will work as long as trait itself is marked pub and there are no plans to change visibility rules that would affect this.

alteous: Okay, that makes sense.

gltf-utils/src/lib.rs
fst3a: Okay, this was moved back in and made use of.

fst3a:
@alteous done as you asked.  I have another question, by the way, should library be more "hardened" as in fail where it can

fst3a:
cc @alteous

gltf-utils/src/casts/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+/// Casting iterator and accompanying types for `TexCoords`.
+pub mod tex_coords;
+/// Casting iterator and accompanying types for `Weights`.
+pub mod weights;
+/// Casting iterator and accompanying types for `MorphWeights`.
+pub mod morph_weights;
+/// Casting iterator and accompanying types for `Rotations`.
+pub mod rotations;
+
+trait Normalizable<T> {
```
alteous: Would prefer [`Normalize`](https://github.com/rust-lang-nursery/api-guidelines/issues/28). Can fix later though.

gltf-utils/src/lib.rs
```diff
@@ -59,147 +42,185 @@ pub trait PrimitiveIterators<'a> {
...
+                (U8, Vec4)  => Colors::RgbaU8(AccessorIter::new(accessor, source)),
+                (U16, Vec4) => Colors::RgbaU16(AccessorIter::new(accessor, source)),
+                (F32, Vec4) => Colors::RgbaF32(AccessorIter::new(accessor, source)),
+                _ => unreachable!(),
             })
     }
 
-    fn indices_u32<'s, S: Source>(&self, source: &'s S) -> Option<IndicesU32<'s>> {
-        self.indices().map(|accessor| IndicesU32(Indices::new(accessor, source)))
+    fn indices<'s, S>(&'a self, source: &'s S) -> Option<Indices<'s>>
```
alteous: Don't worry about the name clash. One can use uniform function call syntax if necessary.

gltf-utils/src/lib.rs
```diff
@@ -337,67 +362,42 @@ impl<T: AccessorItem> AccessorItem for [T; 4] {
...
-
-    /// Morph target animations.
-    Weights(MorphWeightsF32<'a>),
-}
+pub type Inputs<'a> = AccessorIter<'a, f32>;
 
 /// Vertex colors.
-#[derive(Clone, Debug)]
-enum Colors<'a> {
+#[derive(Debug, Clone)]
```
alteous: ?

alteous: C comes before D in the alphabet.

alteous:
It's OK, the new code is more 'complete'. I'm going to try out the new code and report back.

alteous:
> Should methods try to fail with None as much as they can instead of blowing up the thread at => unreachable!() arms, if user chooses to opt out of validation?

The idea behind validation is that the glTF is checked for correctness on behalf of the user. This makes the API much easier to work with. For example, by checking that all indices are in range ahead of time the user doesn't have to deal with returning `Option<TopLevelItem>` for everything. The ability to opt out of validation is intended for cases where the glTF is known to be valid ahead of time so, to answer your question, the library is designed to make working with valid glTF simple and not try to work with invalid glTF.

gltf-utils/src/lib.rs
fst3a: Well, yes, but it's not very nice.  Shadowing Primitive::indices with PrimitiveIterators::indices requires that you to `use` it from a scope higher than Primitive is used at.

gltf-utils/src/casts/mod.rs
fst3a: Alright.

gltf-utils/src/lib.rs
fst3a: I must have accidentally reordered it, sorry.

fst3a:
> The idea behind validation is that the glTF is checked for correctness on behalf of the user. This makes the API much easier to work with. ... to answer your question, the library is designed to make working with valid glTF simple and not try to work with invalid glTF, so panicking is preferred.

Yes, I understand that.  But some methods, such as those found in PrimitiveIterators can fail safely for pretty much free (e.g. by using and_then instead of map).  It's just a slight thought I had, anyway, not really a feature proposal.

fst3a:
cc @alteous

alteous:
I'm going to try this branch out first before merging.

alteous:
This needs to rebased against `incoming` before it can be merged. I'm happy to do it myself but I don't think GitHub will recognise your contributions if I do so. Would you mind if I rebased, squashed, and merged your PR manually?

fst3a:
Sure, go ahead

alteous:
The manual merge has gone horribly wrong! :(

alteous:
I'm going to try and cherry-pick your changes instead.

PR #148, opened 2018-02-23
Fix build issues and bug in sparse accessor parsing

bwasty:
There's a new sample `SimpleSparseAccessor` that failed to import because some optional fields were not marked correctly.
I also fixed the compilation issues as well as I could (without really understanding the related changes.)


alteous:
Thanks!

PR #156, opened 2018-03-12
Implement ExactSizeIterator for every CastingIter

### LINKED ISSUE #155: Missing impl ExactSizeIterator for CastingIter
For example, let's take `gltf_utils::Indices`, which stores an `AccessorIter` that implements ExactSizeIterator. Using `into_u32()` on `Indices` returns an iterator that does not implement `ExactSizeIterator`

alteous:
Fixes #155


alteous:
Fixed by #162

PR #159, opened 2018-04-25
Fix small documentation mistake

magnonellie:
Nothing major, but still, worth fixing IMO. xP

Thank you guys for making this crate!


alteous:
Well spotted. Thank you!

PR #160, opened 2018-05-18
Add some derives for the error type

derekdreery:
Simple 1-line change to add some more impls for the error type.


alteous:
Thanks!

PR #170, opened 2018-06-17
feat: Import from trait to enable non filesystem load scenarios

Rhuagh:


alteous:
Thanks for the pull request.

With regret, I feel compelled to decline this PR since a 'generic' import feature it is out of scope of the crate's goals. The current (0.11.x) `import` feature/function is for convenience and also serves as reference for implementing more complicated import systems. If the `import` function does not solve your problems already then my advice would be to implement your own loader as you/amethyst have done so already.

Note that we had a [similar interface before](https://docs.rs/gltf/0.6.1/gltf/import/trait.Source.html) back in 0.6.x which was heavily criticised.

Is there a particularity compelling reason why this feature should be added?

Rhuagh:
I'm fine with keeping this in amethyst, I just felt it might be nice as a service to other users that need to read from archive files or network like we do.

Issue #173, opened 2018-07-05
Incompatibility with the spec: Glb::to_write serializes Nones as nulls


wlinna:
`Glb::to_write` serializes `None`s as nulls. This is problematic, because gltf specification does not allow `null` values. If a property is set to `None`, it should be omitted.

For example, in the case of `accessors`, the spec doesn't require defining `sparse`, but when it is defined, `null` values are NOT allowed.

Because of this, `GLTFLoader` of ThreeJS throws when I try to use a GLB file serialized by `gltf-rs/gltf`.

Here's a reference to the spec:

https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/accessor.sparse.schema.json

As you can see, `type `is set to `"object"`, not `["object", "null"]`

alteous:
This is quite interesting as I had assumed the [roundtrip_binary_gltf test](https://github.com/gltf-rs/gltf/blob/master/tests/roundtrip_binary_gltf.rs#L60) would have caught this. I imagine this will be an easy fix in gltf-json.

wlinna:
In fact, I've already fixed it locally, but I'll leave the PR for tomorrow

PR #175, opened 2018-07-07
Do not serialize None

wlinna:
Fixes https://github.com/gltf-rs/gltf/issues/173

After these changes (and the now already merged padding fix) I'm able to open serialized binary glTF files in ThreeJS.

Similarly, all empty vectors **must** be left unserialized, but this PR is about `None`s

### DIFF
--- a/gltf-json/src/accessor.rs
+++ b/gltf-json/src/accessor.rs
@@ -126,6 +126,7 @@ pub mod sparse {
 
         /// Optional application specific data.
         #[serde(default)]
+        #[cfg_attr(feature = "extras", serde(skip_serializing_if = "Option::is_none"))]
         pub extras: Extras,
     }
 
@@ -154,6 +155,7 @@ pub mod sparse {
 
         /// Optional application specific data.
         #[serde(default)]
+        #[cfg_attr(feature = "extras", serde(skip_serializing_if = "Option::is_none"))]
         pub extras: Extras,
     }
 
@@ -178,6 +180,7 @@ pub mod sparse {
 
         /// Optional application specific data.
         #[serde(default)]
+        #[cfg_attr(feature = "extras", serde(skip_serializing_if = "Option::is_none"))]
         pub extras: Extras,
     }
 }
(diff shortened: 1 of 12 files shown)


wlinna:
I noticed a problem with this PR. Fix incoming.

wlinna:
I've fixed the problem now

alteous:
I wonder whether there's a less human-error-prone way of doing this, perhaps with a new type?

```rust
#[derive(Deserialize)]
struct VecDeserializeOnly<T>(pub Vec<T>);

impl Serialize for VecDeserializeOnly {
    fn serialize() {
        // nop
    }
}
```

wlinna:
Is your comment by any chance related to [the other PR](https://github.com/gltf-rs/gltf/pull/177)? Or maybe I'm just missing your idea.

Also, why would you turn of serialize?

alteous:
They're very similar cases: the idea is to create a new type whose `Serialize` implementation is specialised in a single place instead of everywhere it is used. The psuedo-code example was intended to demonstrate this idea.

alteous:
Instead of applying `serde(skip_serializing_if)` we `impl Serialize for Extras` such that it is not serialized anywhere.

wlinna:
But my use case **requires** extras. The support for deserializing and serializing `extras` was the entire reason why I picked this library. I used this library for a small utility to write certain metadata to my GLB files.

alteous:
> such that it is not serialized anywhere.

Correction: I mean to say "such that it is serialized as needed", implying the suggested alternative would achieve the same results as the changes in this PR, except the implementation being in one place. Does that make sense? I'll demonstrate with a PR this evening.

wlinna:
Yeah, I think we are on the same page now. I'll be glad to see that PR, as I'm very new to programming in Rust (except that I've read This Week in Rust since its inceptions).

alteous:
After some experimenting I realise my suggestion is not better in any way. Let's go with the simpler option as you have already implemented. Thanks for the PRs, as always! :)

Are there likely to be further changes/PRs in the next couple of days? If so I'd rather wait for them before publishing since since publishing is time consuming.

wlinna:
I'm not going to make any PRs unless I notice something new, which is unlikely, given that the library already works well for my use case.

You're welcome :)

alteous:
I wonder whether we need to address `HashMap` too.

alteous:
Hmm... `attributes` is a required field so the deserialization should fail. If the user happens to remove all the attributes and attempts to serialize then perhaps the serializer should fail.

PR #183, opened 2018-07-15
Skip serializing extras::Void's field

Ralith:
Fixes injection of an extraneous field into generated JSON.

It'd be nice to skip Extras entirely without going through and tediously cfg-ing every use, but I didn't find a way to do that.


alteous:
> It'd be nice to skip Extras entirely without going through and tediously cfg-ing every use, but I didn't find a way to do that.

This is the rationale for having `Extras = Void`. The alternative is, as you state, to stick `#[cfg(feature = "extras")]` everywhere.

PR #181, opened 2018-07-15
Implement KHR_materials_pbrSpecularGlossiness

Ralith:
Refer to [the spec](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness/README.md). Not yet tested on real assets; seeking initial feedback.

PR #184, opened 2018-07-15
Propagate detailed error information in Display impl

Ralith:
This would have made my debugging of #181 (and a few other issues) go much smoother.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -457,8 +457,23 @@ impl Document {
 
 impl std::fmt::Display for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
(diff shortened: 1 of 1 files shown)


alteous:
I don't understand the issue here. Could you detail exactly why "JSON error" was reported on the duck model (I suspect the validation steps failed) and whether `println!("{:?}", error)` was more detailed?

alteous:
I'm unable to reproduce the issue described in #181 because the code has been fixed since. :)

Ralith:
The really important change here is just forwarding to the wrapped errors' `Display` impls instead of ignoring them.

alteous:
Seems reasonable to me. Thanks. :)

Issue #189, opened 2018-08-28
gltf_json::Index cannot be constructed by user code.


anderejd:
`gltf_json::Index` cannot be constructed by user code since the constructor and the members are private.

This prevents this library to be used for implementing glTF export. Would you be Ok with making it public? EDIT: (the constructor and/or the members)

Since buffers imported from external glTF data can be modified by the user code, the Index type does not provide protection against indexing errors. Maybe I'm not understanding the intention of the members of the Index type being private?

anderejd:
By the way, I have more modifications coming to improve serialization/export support. Would you prefer that I push all of them to this PR or open multiple small ones instead?

anderejd:
I'm going to go ahead and close this and open a larger PR that contains all changes required to implement export, I hope this is ok :slightly_smiling_face:

anderejd:
Ok I've not had a proper dose of coffee yet, this issue is not closed, I intended to close the PR, not this issue. Sorry for the noise! Re-opening.

alteous:
> By the way, I have more modifications coming to improve serialization/export support. Would you prefer that I push all of them to this PR or open multiple small ones instead?

Whatever is easiest for you. I would prefer a single PR containing incremental commits because that's easy to review and keeps the git history useful.

PR #191, opened 2018-08-31
Changes to support glTF 2.0 export/serialization

anderejd:
Solves: #189 

Please see the commit messages for the details related to each commit.

Some background:
I'm implementing glTF 2.0 export based on this crate, not export of imported data, export of in-process generated assets. This requires all related data structures to be correctly instantiated by user code. I have also made optional json properties optional and added `serde(skip_serializing_if` attributes for some simple default values. This helps to verify that my test output is identical to the reference models provided by Khronos. The TRS properties also must be made optional to follow the specification, please see the commit messages and links to the related sections of the spec.

I have started to implement some builders to make export easier, that aren't ready yet, but I would like to upstream them as well at a later time, they do depend on this PR.

If you would like me to make further changes or go about this in a completely different way, please tell me :slightly_smiling_face:

### DIFF
--- a/gltf-json/src/accessor.rs
+++ b/gltf-json/src/accessor.rs
@@ -232,7 +232,7 @@ pub struct Accessor {
     pub name: Option<String>,
 
     /// Specifies whether integer data values should be normalized.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "is_normalized_default")]
     pub normalized: bool,
     
     /// Sparse storage of attributes that deviate from their initialization
@@ -240,6 +240,11 @@ pub struct Accessor {
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub sparse: Option<sparse::Sparse>,
+}
+
+// Help serde avoid serializing this glTF 2.0 default value.
+fn is_normalized_default(b: &bool) -> bool {
+    *b == bool::default()
 }
 
 /// The data type of an index.
--- a/gltf-json/src/accessor.rs
+++ b/gltf-json/src/accessor.rs
@@ -121,8 +121,8 @@ pub mod sparse {
         pub component_type: Checked<IndexComponentType>,
 
         /// Extension specific data.
-        #[serde(default)]
-        pub extensions: extensions::accessor::sparse::Indices,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        pub extensions: Option<extensions::accessor::sparse::Indices>,
 
         /// Optional application specific data.
         #[serde(default)]
@@ -150,8 +150,8 @@ pub mod sparse {
         pub values: Values,
 
         /// Extension specific data.
-        #[serde(default)]
-        pub extensions: extensions::accessor::sparse::Sparse,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        pub extensions: Option<extensions::accessor::sparse::Sparse>,
 
         /// Optional application specific data.
         #[serde(default)]
@@ -175,8 +175,8 @@ pub mod sparse {
         pub byte_offset: u32,
 
         /// Extension specific data.
-        #[serde(default)]
-        pub extensions: extensions::accessor::sparse::Values,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        pub extensions: Option<extensions::accessor::sparse::Values>,
 
         /// Optional application specific data.
         #[serde(default)]
@@ -205,8 +205,8 @@ pub struct Accessor {
     pub component_type: Checked<GenericComponentType>,
 
     /// Extension specific data.
-    #[serde(default)]
-    pub extensions: extensions::accessor::Accessor,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub extensions: Option<extensions::accessor::Accessor>,
 
     /// Optional application specific data.
     #[serde(default)]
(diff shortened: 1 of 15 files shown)


gltf-json/src/buffer.rs
```diff
@@ -100,15 +103,21 @@ pub struct View {
...
+    pub extensions: Option<extensions::buffer::View>,
 
     /// Optional application specific data.
     #[serde(default)]
     #[cfg_attr(feature = "extras", serde(skip_serializing_if = "Option::is_none"))]
     pub extras: Extras,
 }
 
+impl View {
+    pub fn byte_offset_default() -> u32 {
```
alteous: What is this for?

gltf-json/src/root.rs
```diff
@@ -182,7 +182,7 @@ impl Root {
 
 impl<T> Index<T> {
     /// Creates a new `Index` representing an offset into an array containing `T`.
-    fn new(value: u32) -> Self {
+    pub fn new(value: u32) -> Self {
```
alteous: :+1:

src/buffer.rs
```diff
@@ -137,7 +137,8 @@ impl<'a> View<'a> {
 
     /// Returns the offset into the parent buffer in bytes.
     pub fn offset(&self) -> usize {
-        self.json.byte_offset as usize
+        self.json.byte_offset.unwrap_or_else(
+            json::buffer::View::byte_offset_default) as usize
```
alteous: Unless there is a compelling reason to do so otherwise, I would prefer the previous way.

src/scene/mod.rs
```diff
@@ -176,9 +176,12 @@ impl<'a> Node<'a> {
             }
         } else {
             Transform::Decomposed {
-                translation: self.json.translation,
-                rotation: self.json.rotation.0,
-                scale: self.json.scale,
+                translation: self.json.translation.unwrap_or_else(
```
alteous: I suppose this is to make it easier to specify default transformation properties? If so, consider making a new type and implementing `Default::default` on it. Then the user can use `Foo { bar: 123, .. Default::default() }` to populate the rest of the struct with default values.

alteous:
Looks good to me generally. I'd like to discuss a bit more before submitting.

P.S: Sorry for lack of activity lately; I've been very busy at work this week.

gltf-json/src/accessor.rs
```diff
@@ -242,6 +242,11 @@ pub struct Accessor {
     pub sparse: Option<sparse::Sparse>,
 }
 
+// Help serde avoid serializing this glTF 2.0 default value.
+fn is_normalized_default(b: &bool) -> bool {
+    *b == bool::default()
```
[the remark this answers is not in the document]
anderejd: Yeah this looks like I made a brainfart. ><

Would you prefer `*b` or `*b == false`?

anderejd:
Thanks for the review. 👍 I'll be back at a computer tomorrow, on mobile now, will reply to your review tomorrow.

gltf-json/src/buffer.rs
anderejd: This is called from `src/buffer.rs`:
```rust
    /// Returns the offset into the parent buffer in bytes.
    pub fn offset(&self) -> usize {
        self.json.byte_offset as usize
        self.json.byte_offset.unwrap_or_else(
            json::buffer::View::byte_offset_default) as usize
```
You posted a review comment regarding this call site here:
https://github.com/gltf-rs/gltf/pull/191#discussion-diff-214507143R141

The intention here was to give the default value a descriptive name instead of inserting the literal `0` in various places.

src/scene/mod.rs
anderejd: Short answer, to follow the glTF 2.0 specification.

Longer answer, this change is a result of me changing the TRS properties to be optional to follow the specification. There's also a practical reason for this change, to allow exporting a node with the `.matrix` property present since the TRS properties and the `.matrix` properties are mutually exclusive. And any combination of `.translation`, `.rotation` and `.scale` is allowed to be present, if `.matrix` is not.

> A node can have either a matrix or any combination of translation/rotation/scale (TRS) properties.
<https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#node>

anderejd:
I think I've replied to your comments and questions in the review now. 

> P.S: Sorry for lack of activity lately; I've been very busy at work this week.

No worries, I didn't see any suggestion of free commercial support in the README :smile:

gltf-json/src/accessor.rs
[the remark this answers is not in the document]
alteous: This is testing if `b` is `false`, right? If so, then `!(*b)`.

gltf-json/src/scene.rs
```diff
@@ -66,16 +75,26 @@ pub struct Node {
     pub weights: Option<Vec<f32>>,
 }
 
-fn node_scale_default() -> [f32; 3] {
-    [1.0, 1.0, 1.0]
+impl Node {
```
alteous: I'd rather these default values be written inline for the same reasons as `byte_offset`.

src/scene/mod.rs
alteous: Understood. This is a nice improvement for not writing default values.

Regarding the builders: It would probably be easier to sort out the JSON first. An early/WIP review would be good.

anderejd:
All remarks as I understood them should be fixed by: https://github.com/gltf-rs/gltf/pull/191/commits/5c0b8061e64057680bd411020a47b9d5fe59a11c

alteous:
Good job, thanks!

PR #192, opened 2018-09-08
Add slice import for web targets

ForsakenHarmony:
will crash in wasm on gltf files with external blobs, but that's to be expected

### DIFF
--- a/src/import.rs
+++ b/src/import.rs
@@ -163,11 +163,7 @@ pub fn import_image_data(
     Ok(images)
 }
 
-fn import_impl(path: &Path) -> Result<Import> {
-    let base = path.parent().unwrap_or(Path::new("./"));
-    let file = fs::File::open(path).map_err(Error::Io)?;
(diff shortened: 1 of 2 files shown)


alteous:
Thanks for your pull request.

Unfortunately, I'm unwilling to merge as-is. What bothers me is this line:
```rust
import_impl(Gltf::from_slice(slice)?, Path::new("./"))`
```
What this does it assumes any relative URIs are relative to the current working directory, which I think is unreasonable.

ForsakenHarmony:
The point is that there is no file system on the web, implementing async loading of other files would be a huge undertaking

ForsakenHarmony:
OK?

ForsakenHarmony:
I guess you're just never gonna explain why you just closed this?

ForsakenHarmony:
Or is this possible with `Gltf::from_slice`

ForsakenHarmony:
Nope

Issue #187, opened 2018-08-20
Export (GltfBuilder)


anderejd:
I am going to do some sort of gltf export in the near future and I would like it to be based on this library. Is export functionality on the road map? Would a PR be welcome?

I found this (part of a) comment in another thread:

> Being about to export glTF would be a great future feature. I'm considering a builder struct GltfBuilder where we could make guarantees of the validity of references the users provides, but personally I'd rather concentrate on just loading at the moment.

alteous:
The builder idea is not on the 1.0 roadmap. Pull requests are always welcome.

The crate supports exporting binary glTF and serialising the JSON for standard glTF.

alteous:
I should mention: There is no specific 1.0 roadmap but I am keen to release a version 1.0 shortly as the API is stable. Extensions are the last thing to be addressed.

anderejd:
Thanks for the reply!

> The builder idea is not on the 1.0 roadmap. Pull requests are always welcome.

I might look at that tomorrow. At this point the only thing I need is geometry export for some initial tests. If this is too much work to implement in this crate my fallback plan is to simply use the triangle sample model file as a template and modify the needed parts to export my data instead.

> The crate supports exporting binary glTF and serialising the JSON for standard glTF.

I searched for "export" in the docs and the code and didn't find anything. I also looked for builders and constructors, for example on Mesh the constructor is internal: https://docs.rs/gltf/0.11.2/src/gltf/mesh/mod.rs.html#134 That made me think that the entire in-memory structure is meant to be read-only after it has been created by calling the import functions. What I need is to build the gltf document in-memory, and then serialize it to a file.

Can you give an example of how one can use this crate to build the gltf document in-memory and then export it? What am I missing?

alteous:
> the entire in-memory structure is meant to be read-only after it has been created by calling the import functions

This is correct.  The `Document` struct is the deserialised form of the glTF JSON. Almost all other structs are wrappers around `Document` enabling you to traverse the scene hierarchy easily.

There is no nice way of building/exporting glTF at the moment; however, you are free to call `Document::{from_json, into_json}` and to manipulate the JSON as you wish. The builder idea aims to do exactly this but in a friendly way.

Side note: the `gltf-json` is considered an implementation detail and is inherently unstable.

anderejd:
> Side note: the gltf-json is considered an implementation detail and is inherently unstable.

After a quick look at the gltf-json crate, it seems like what I'm after, thanks!

[...]

PR #193, opened 2018-09-10
Export builders and example

anderejd:
## This PR
This is an initial attempt at improving export ergonomics and correctness. It's intended to help with the most basic export needs, but should be easy to extended later.

## Future work/PRs
 - More builders and component types to handle more advanced scenes.
 - Adding another level of builders on top of these builders could be interesting to allow all of buffer data, buffers, buffer views and accessors to be built by the same builder.

## Related to
 - #187

### DIFF
--- a/gltf-json/src/builders.rs
+++ b/gltf-json/src/builders.rs
@@ -1,0 +1,1073 @@
+//! Builders indended to help with gltf export implementation.
+//!
+//! NOTE: Using consuming builders for simplicity. Change to non-consuming
+//! builders if benchmarks prove it necessary.
+//!
+
+use byteorder::{ByteOrder, LittleEndian};
+use ::accessor::sparse::Sparse;
+use ::accessor::{GenericComponentType, Type};
+use ::buffer::{Buffer, ByteStride, Target, View};
+use ::camera::Camera;
+use ::extensions;
+use ::material::Material;
+use ::mesh::{Mode, MorphTarget, Primitive, Semantic};
+use ::scene::UnitQuaternion;
+use ::skin::Skin;
+use ::validation::Checked;
+use ::validation::Checked::Valid;
+use ::{Accessor, Extras, Index, Mesh, Node, Scene};
+use num_traits::cast::ToPrimitive;
+use serde_json;
+use std::collections::HashMap;
+use std::io;
+
+/// A node in the node hierarchy.  When the node contains `skin`, all
+/// `mesh.primitives` must contain `JOINTS_0` and `WEIGHTS_0` attributes.
+/// A node can have either a `matrix` or any combination of
+/// `translation`/`rotation`/`scale` (TRS) properties. TRS properties are converted
+/// to matrices and postmultiplied in the `T * R * S` order to compose the
+/// transformation matrix; first the scale is applied to the vertices, then the
+/// rotation, and then the translation. If none are provided, the transform is the
+/// identity. When a node is targeted for animation (referenced by an
+/// animation.channel.target), only TRS properties may be present; `matrix` will not
+/// be present.
+///
+/// Can be removed if Node implements Default.
+/// Should we keep it for API consistency?
+///
+#[derive(Default)]
+pub struct NodeBuilder {
+    camera: Option<Index<Camera>>,
+    children: Option<Vec<Index<Node>>>,
+    extensions: Option<extensions::scene::Node>,
+    extras: Extras,
+    matrix: Option<[f32; 16]>,
+    mesh: Option<Index<Mesh>>,
+
+    #[cfg(feature = "names")]
+    name: Option<String>,
+
+    rotation: Option<UnitQuaternion>,
+    scale: Option<[f32; 3]>,
+    translation: Option<[f32; 3]>,
+    skin: Option<Index<Skin>>,
+    weights: Option<Vec<f32>>,
+}
+
+impl NodeBuilder {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// The index of the camera referenced by this node.
+    pub fn camera(mut self, c: Index<Camera>) -> Self {
+        self.camera = Some(c);
+        self
+    }
+
+    /// The indices of this node's children.
+    pub fn children(mut self, c: Vec<Index<Node>>) -> Self {
+        self.children = Some(c);
+        self
+    }
+
+    /// Extension specific data.
+    pub fn extensions(mut self, e: extensions::scene::Node) -> Self {
+        self.extensions = Some(e);
+        self
+    }
+
+    /// Optional application specific data.
+    pub fn extras(mut self, e: Extras) -> Self {
+        self.extras = e;
+        self
+    }
+
+    /// 4x4 column-major transformation matrix.
+    ///
+    /// TODO: Use session types to divide the control flow for .matrix and
+    /// decomposed mode.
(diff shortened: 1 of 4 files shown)


anderejd:
Manual benchmarks looks promising, building a test mesh of 10 000 000 indexed triangles and exporting it to gltf with external binary buffer on disk takes about 1 second on my machine, the export looks like it's maybe half of the total time. It's way better than I was hoping for at this stage :slightly_smiling_face:

anderejd:
Rebased and squashed the corrections made since this PR was opened. I think this is ready for review now.

anderejd:
Sorry in advance for the spam.

Should this and everything related to export be put behind a cargo feature flag or perhaps moved to and published in a new crate? I imagine that only a tiny part of users will be interested in export and moving forward with my project I will be wanting to flesh out the builders with more code. I think a feature flag would be a nice option in case the export helper code would need to access to internals later.

Regarding feature flags, would it make sense to remove the current feature flags and make that functionality standard? From what I have seen the current feature flags enable a few optional properties and don't really change much for the user code? At least the "extras" and "names" seems like they would make sense to remove and have that code enabled in all builds to avoid all the `#[cfg(feature = ` attributes.

gltf-json/src/builders.rs
```diff
@@ -0,0 +1,1073 @@
...
+    pub min: Option<serde_json::Value>,
+
+    /// Maximum value of each component in this attribute.
+    /// POSITION accessor must have min and max properties defined.
+    pub max: Option<serde_json::Value>,
+}
+
+impl AccessorBuilder {
+    /// Advance to the optional stage of this builder.
+    pub fn with_options(self) -> AccessorBuilderWithOptions {
```
alteous: Is it possible to not have to call `with_options` like `PrimitiveBuilder`?

alteous:
Wow, thank you very much! This is a great start.

I have a lot of catching up to do. Let me start by answering your questions:

> Should this and everything related to export be put behind a cargo feature flag or perhaps moved to and published in a new crate?

Please, not a new crate. :) I'd much prefer a cargo feature flag as they are far easier to maintain than separate crates. Believe me: we've tried this before!

> Regarding feature flags, would it make sense to remove the current feature flags and make that functionality standard?

Perhaps. On Gitter and IRC people seem to prefer lean libraries with opt-in features rather than fat libraries with opt-out features. The actual reason why they are opt-out currently is because they are relatively expensive to capture due to being owned and that they are not strictly necessary for an application to import.

Regarding the design, I have two concerns:

1. I don't like the `with_options` function; please revise this.
2. Is it possible to export interleaved vertex data in a 'safe' way? I envisioned a `#[derive(Vertex)]`-like thingamajig that would generate the accessor field values automatically.

anderejd:
Thanks for the feedback! 

> Please, not a new crate. :) I'd much prefer a cargo feature flag as they are far easier to maintain than separate crates. Believe me: we've tried this before!

Sure thing, feature flag it is! I'll hide all the new code behind a new cargo feature and I'll also leave the current cargo feature flags as is.

> I don't like the with_options function; please revise this.

Yeah... I'm not entirely happy with that one either.  The primary purpose of the `with_options` thingy is to force named parameters both for required and optional fields, while avoiding copy pasting `: default::Default()` for every optional field in the user code. I think named parameters/fields are important when there are many parameters of the same type or just similar numeric types.
 
I considered marker types like this: <https://dev.to/mindflavor/rust-builder-pattern-with-types-3chf> but that seems to me to add more complexity than it's worth(?).

Another simple option would be to move the required parameters to a struct called something like `SomethingBuilderRequiredParams` and take that as a required field on the actual builder?

> Is it possible to export interleaved vertex data in a 'safe' way? I envisioned a #[derive(Vertex)]-like thingamajig that would generate the accessor field values automatically.

I haven't put much thought into interleaved formats yet beyond whats provided in the comments. At this stage I'm leaning towards the feeling that interleaved formats will require a completely different code path with its own dedicated builder and `Read` implementation to avoid unnecessary growth of code complexity.

gltf-json/src/builders.rs
anderejd: Yes, the easiest way would be to drop compile time checks for required vs optional parameters and return a result from .build. Please see my reply https://github.com/gltf-rs/gltf/pull/193#issuecomment-421555650 for some follow-up questions.

anderejd:
Closing due to inactivity.

I'm moving this code back inside my project and will develop it there until ready. I want support for Iterator instead of slices to avoid large extra allocations on export as well as support for interleaved formats. A higher level export builder would also be nice. This is likely going to take quite some time until I'm satisfied with the overall shape of things.

I'll open a new PR, when it's done.

anderejd:
Should we reopen this?

alteous:
Yes, I have a weekend free to work on this. :)

alteous:
The interesting case is being able to prepare buffer data for export and I'm not sure how to approach this.

Generating builders for the json data structures is straightforward. I'm experimenting with the derive_builder crate for this and that seems to work well.

anderejd:
I'm going to be AFK for most of this weekend so I'll try to make a proper braindump in this reply.

>The interesting case is being able to prepare buffer data for export and I'm not sure how to approach this.

The way I went about this for this PR was to work towards a specific use case: exporting high resolution triangle meshes. The example in this PR only exports a single triangle but I can provide a meshlite based high-res export example if you think that would be useful, it will look almost identical to the existing example but with some meshlite glue code added.

To push this PR forward I would add more examples of increasing complexity to drive the API design, perhaps using three-rs as a export data source for some of the more advanced examples.

We did start some initial code review a couple of months ago in this PR, I'm not sure how you want to resolve that, feel free to modify this PR as you see fit. I'm pretty pleased with how the API looks and works for the current export example, but I suspect that it'll need to evolve when adding more complex export examples. If you feel unsure about evolving the export API inside gltf-rs due to API churn, we could just as well do this in a new crate and let that crate stay in 0.x mode until it's ready.

To summarize, I would recommend adding more examples of increasing complexity and build out the export API based on the needs of the examples.

Hope that helps, I'll have more time next week. Have a nice (and productive!) weekend :)

aloucks:
This looks promising! Are there plans to merge this in the near term?

One issue is that the `Primitive` attributes are stored in a `HashMap` with `RandomState`. The serialized output won't be deterministic given the same input. I think it would be better to change this to use `BTreeMap` or use a deterministic hash state (e.g. `FnvBuildHasher`).

anderejd:
> One issue is that the `Primitive` attributes are stored in a `HashMap` with `RandomState`. The serialized output won't be deterministic given the same input. I think it would be better to change this to use `BTreeMap` or use a deterministic hash state (e.g. `FnvBuildHasher`).

Good point! Let's open an issue to change that in Primitives as well.

> This looks promising! Are there plans to merge this in the near term?

I would like to push this PR to an acceptable state for merging but it seems like I'm waiting for more feedback while @alteous could be waiting for me to provide more code 😄 

I want to push this forward, how would you like to proceed? Review and change this PR or close and move it to a new crate? Please don't hesitate to close it if you think this is going in the completely wrong direction to be merged in this crate.

bwasty:
If @alteous doesn't find the time, I'll try to have a proper look at this soonish.

alteous:
@anderejd : sorry but I want to move the crate in a different direction. See #234 for the details. Thank you for your work nonetheless.

anderejd:
No problem :)

PR #205, opened 2019-02-24
Fix assert on meshes on byte_stride = 0

repi:
Have found glTF meshes (on Google Poly) that specify 0 byte stride for index buffers which caused debug_assert to be triggered on 0 byte stride when calling `read_indices`.

This should already fail validation, but only the case of `validation_completely` which is typically not run.

With this change we specifically handle the case of byte stride 0 and treat it as if no byte stride was specified to avoid the debug assert and panicking.


alteous:
Thanks!

PR #206, opened 2019-03-07
Update to 2018 edition

aleksijuvani:
Updates the crate to use the 2018 edition of Rust and fixes all resulting compile errors. This will allow the crate to take advantage of new features such as non-lexical lifetimes in new code.


alteous:
Thanks for the PR!

Does this change the supported compiler versions? Please amend the README if so since currently it says 'rustc version 1.19 or above is required; version 1.26 and above is recommended'.

aleksijuvani:
I believe it does. I think the minimum requirement is 1.31, but I will take a better look at this tomorrow and amend the pull request accordingly.

aleksijuvani:
Done.

Issue #208, opened 2019-03-08
Index<T> is not copy for Non-Copy T


Veykril:
I just noticed that you have to clone Indices manually instead of being able to rely on implicit copying for almost if not all gltf objects due to how derives work for `Copy` and `Clone` and almost all objects not being `Copy` due to their fields. 
https://github.com/gltf-rs/gltf/blob/33197d6231c900ffcef5f196bfa1e258dcc0c310/gltf-json/src/root.rs#L20-L21

Is there a specific reason for this or was this just overlooked? To my understanding an `Index` doesnt own the `T` anyways, so there shouldn't be an issue with implementing those two traits manually for all `T` instead, and changing the `PhantomData<T>` to `PhantomData<*const T>` to hint the compiler that it is not owned by the struct. The latter part would require to manually implement `Send` and `Sync` for `Index<T>` as well though since it still is actually thread-safe.

alteous:
> Is there a specific reason for this or was this just overlooked?

Likely overlooked. Previous versions of Rust may have disallowed `Copy` for `PhantomData` but I can't remember.

alteous:
Fixed in #209.

PR #209, opened 2019-03-08
Make Index<T> Copy for all T

Veykril:
The fix mentioned in my issue #208, this should more clearly show what I mean.

I also noticed making it `Copy` makes it possible to just pass it by value instead, given that it is effectively just an `u32`.


alteous:
Thanks!

PR #210, opened 2019-03-13
Fix accessor performance

### LINKED ISSUE #207: Import performance
I'm trying to import a model of the Sponza Atrium, but I'm experiencing some really atrocious import performance. The model has over 300 meshes which results in over 1500 accessors in the glTF file. Importing the model takes several seconds in release mode and an order of magnitude longer in debug mode (it took so long I didn't even bother measuring it).

Profiling seems to indicate that 74 % (or 93 % with a binary model) of the import time is spent inside `gltf::accessor::Accessor::new` and 15 % (18 % with a binary model) is spent inside `gltf::buffer::View::new`.

aleksijuvani:
Fixes #207. Fixing the issue turned out to be deceptively simple. It was only a matter of removing the unused `view` field from the `Accessor` struct. Previously, reading in the Sponza Atrium model took up to 5 seconds, but with this PR only takes up around 55 milliseconds on my system.

### DIFF
--- a/src/accessor/mod.rs
+++ b/src/accessor/mod.rs
@@ -85,9 +85,6 @@ pub struct Accessor<'a> {
 
     /// The corresponding JSON struct.
     json: &'a json::accessor::Accessor,
-
-    /// The buffer view this accessor reads from.
-    view: buffer::View<'a>,
 }
 
 impl<'a> Accessor<'a> {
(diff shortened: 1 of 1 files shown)


alteous:
Wow this wasn't even used... :blush:

PR #212, opened 2019-03-31
Upgrade Dependencies

nuew:
This PR updates the following outdated dependencies, and fixes the resulting incompatibilities:

- `approx`, 0.1.1 -> 0.3
- `base64`, 0.6 -> 0.10
- `cgmath`, 0.15 -> 0.17
- `image`, 0.19 -> 0.21
- `lazy_static`, 0.2 -> ^1
- `quote`, 0.3 -> 0.6
- `syn`, 0.11 -> 0.15

Updating `quote` requires adding a dependency on `proc_macro2` to `gltf-derive`, but `gltf-json` already transitively requires that anyways.

Updating `image` is slightly complicated by their adding of BGR pixel formats to the `image::DynamicImage` enum. The current PR just adds the new formats to `gltf::image::Format`, though you could, to preserve backwards compatibility, alternatively not add them and report some sort of error when encountered; or, to ensure forwards compatibility, add a hidden __Nonexhaustive variant (see [RFC 2008](https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md) to prevent exhaustive matching.

Additionally, this only asks for `image` features required by the glTF spec (that is, support for PNG and JPEG images), so as to decrease unused code size if someone is not using `image` except through `gltf`. `image`s `jpeg_rayon` parallel JPEG decoding feature is also enabled by default for convenience. Alternatively, of course, we could not do this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,8 +26,13 @@ base64 = { optional = true, version = "0.6" }
 byteorder = "1.1"
 cgmath = "0.15"
 gltf-json = { path = "gltf-json", version = "0.11.3" }
-image = { optional = true, version = "0.19" }
 lazy_static = "0.2"
+
+[dependencies.image]
+default-features = false
+features = ["jpeg", "png_codec"]
+optional = true
+version = "0.21"
 
 [features]
 default = ["import", "utils", "names"]
@@ -36,6 +41,7 @@ names = ["gltf-json/names"]
 utils = []
 import = ["base64", "image"]
 KHR_materials_pbrSpecularGlossiness = ["gltf-json/KHR_materials_pbrSpecularGlossiness"]
+image_jpeg_rayon = ["image/jpeg_rayon"]
 
 [[example]]
 name = "gltf-display"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,14 +19,14 @@ travis-ci = { repository = "gltf-rs/gltf" }
 members = ["gltf-derive", "gltf-json"]
 
 [dev-dependencies]
-approx = "0.1.1"
+approx = "0.3"
 
 [dependencies]
-base64 = { optional = true, version = "0.6" }
+base64 = { optional = true, version = "0.10" }
 byteorder = "1.1"
-cgmath = "0.15"
+cgmath = "0.17"
 gltf-json = { path = "gltf-json", version = "0.11.3" }
-lazy_static = "0.2"
+lazy_static = "1"
 
 [dependencies.image]
 default-features = false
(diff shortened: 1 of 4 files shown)


gltf-derive/src/lib.rs
```diff
@@ -4,31 +4,31 @@
...
     };
     let ident = &ast.ident;
-    let minimal_validations: Vec<quote::Tokens> = fields.iter()
+    let minimal_validations: Vec<TokenStream> = fields
+        .iter()
         .map(|f| f.ident.as_ref().unwrap())
         .map(|ident| {
             use inflections::Inflect;
-            let field = ident.as_ref().to_camel_case();
+            let field = ident.to_string().to_camel_case();
```
alteous: Why is `to_string` needed?

Cargo.toml
```diff
@@ -19,23 +19,29 @@ travis-ci = { repository = "gltf-rs/gltf" }
...
-cgmath = "0.15"
+cgmath = "0.17"
 gltf-json = { path = "gltf-json", version = "0.11.3" }
-image = { optional = true, version = "0.19" }
-lazy_static = "0.2"
+lazy_static = "1"
+
+[dependencies.image]
+default-features = false
+features = ["jpeg", "png_codec"]
```
alteous: This is better.

alteous:
Thanks for your PR!

The changes to the image crate are good. Adding `#[non_exhaustive]` won't be necessary.

The parallel JPEG decoding feature is OK but I would prefer it to be opt-in, that is, not enabled by default.

gltf-derive/src/lib.rs
nuew: Using .as_ref() to get a `str` was [removed in syn 0.13](https://github.com/dtolnay/syn/commit/a74a1c89fd76b4f2ba09f6dddf916287b0565a50#diff-2e7b2e9a81356f6d913575ff00a00e87), because [proc_macro2 0.4 removed](https://github.com/alexcrichton/proc-macro2/commit/f388843b8a92b5523c30d51f1da436a05ebb5551#diff-b4aea3e418ccdb71239b96952d9cddb6) their `Ident` `AsStr` impl, as the normal proc_macro crate removed their `AsStr` impl, the rationale for which is explained [here](https://github.com/rust-lang/rust/issues/38356#issuecomment-386786645).

nuew:
Parallel JPEG decoding was only enabled by default because it had been in `image`, though I think all of their features are enabled by default.

I also fixed the alphabetization on the third commit message.

PR #221, opened 2019-05-01
Add sRGB color space docs to base_color_texture

aloucks:
The emissive and specular_glossiness texture docs explicitly noted that the values are in sRGB color space, but was missing from base_color_texture. 

https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#pbrmetallicroughnessbasecolortexture

### DIFF
--- a/src/material.rs
+++ b/src/material.rs
@@ -191,7 +191,8 @@ impl<'a> PbrMetallicRoughness<'a> {
         self.json.base_color_factor.0
     }
 
-    /// Returns the base color texture.
+    /// Returns the base color texture. The texture contians RGB(A) components
+    /// in sRGB color space.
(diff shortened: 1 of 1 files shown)


src/material.rs
```diff
@@ -191,7 +191,8 @@ impl<'a> PbrMetallicRoughness<'a> {
         self.json.base_color_factor.0
     }
 
-    /// Returns the base color texture.
+    /// Returns the base color texture. The texture contians RGB(A) components
```
alteous: contains

alteous:
Hi, thanks for the pull request!

There's one typo. I can fix it for you before pushing if you like.

src/material.rs
aloucks: Fixed!

alteous:
Thank you!

PR #224, opened 2019-06-24
Use nalgebra instead of cgmath

### LINKED ISSUE #197: Replace cgmath by nalgebra through feature flags
Hello,

From what I gathered, gltf uses cgmath for pure operations over Vectors/Matrices/Quaternions (in scene::Transform);
It'd be nice to have an optional dependency on cgmath for those who are using nalgebra in their projects instead
I can try starting a PR if this is something you'd be willing to merge

Thanks

alteous:
Closes #197

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The `gltf` crate adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - Renamed `validate_minimally` as `validate`.
+- Internal transform conversions now use `nalgebra` instead of `cgmath`.
 
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,9 @@ approx = "0.3"
 [dependencies]
 base64 = { optional = true, version = "0.10" }
 byteorder = "1.1"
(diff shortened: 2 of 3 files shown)


aloucks:
Is this really that beneficial? nalgebra takes 28 seconds to build on my machine. cgmath takes 5 seconds.

alteous:
It seems `cgmath` is no longer actively maintained and people are switching to `nalgebra` instead. If people are already using `nalgebra` in their projects then I imagine there will be no increase in build time.

aloucks:
Not everyone is switching to nalgebra right now and it's a big dependency for a handful of transformation functions. Would you be open to having a `math` module in gltf-rs that adds the transformation functions from cgmath or nalgebra (and dropping both dependencies)?

alteous:
#223 implements what you describe. The code was taken from `cgmath` so there is a licensing concern: can we sub-licence as MIT?

alteous:
Web search reveals Apache is not a copyleft licence, which means derivative works or modified versions are not required to be released under the same licence. The restriction is that the original licence must be retained and identified as having been modified. Therefore #223 should be a solution.

alteous:
Closed in favour of #223.

PR #244, opened 2019-08-05
separate import from loading data

### LINKED ISSUE #243: Separate image+base64 vs. import
There are two steps to loading images:

1. Parse the data, and determine if it's a uri, raw bytes, etc.
2. Load the results of that into a proper image

Step 1 is totally internal to the crate and doesn't have any dependencies. Makes sense that this is behind the "import" feature.

However, step 2 has external dependencies, and in some environments (e.g. the web), loading the data would be done some other way.

Can I request that anything which depends on `image_crate` be put behind some other feature flag, perhaps "import-image-data" ? (the import_image_data function itself, from import, could be put behind this as well)

dakom:
fixes #243

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,11 +34,12 @@ optional = true
 version = "0.21"
 
 [features]
-default = ["import", "utils", "names"]
+default = ["import", "import_data_reference", "utils", "names"]
 extras = ["gltf-json/extras"]
 names = ["gltf-json/names"]
 utils = []
-import = ["base64", "image"]
+import = []
+import_data_reference = ["base64", "image"]
 KHR_lights_punctual = ["gltf-json/KHR_lights_punctual"]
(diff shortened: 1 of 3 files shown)


dakom:
Doesn't seem like the AppVeyor build failing has anything to do with the code... any ideas?

alteous:
Thanks for the PR. The AppVeyor build can be ignored as it's not setup yet.

To confirm: there is no functional change, right?

dakom:
Right - for this PR all I did was separate out the image/base64/import helpers into a new feature called `import_data_reference`, and even with that, it's added to the default so there should be no change

dakom:
Thanks for merging it! 

FYI I didn't bump the version number, so if you have a continuous deployment setup, it'll need to be bumped first :)

alteous:
I don't bump the version number until the last minute. The CI setup clones directly from master.

PR #249, opened 2019-08-28
Give gltf::material::Material Clone & Debug derives.

### LINKED ISSUE #248: gltf::material::Material has no derives
Hi folks – curious if the lack of Clone/Debug on Material is deliberate, or a simple omission?

https://github.com/gltf-rs/gltf/blob/master/src/material.rs#L10

This does make it more cumbersome to do transformation from one glTF files to another...

zellski:
Fixes #248.

### DIFF
--- a/src/material.rs
+++ b/src/material.rs
@@ -7,6 +7,7 @@ lazy_static! {
 }
 
 /// The material appearance of a primitive.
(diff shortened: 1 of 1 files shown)


alteous:
Thanks!

PR #254, opened 2019-09-30
Add slice import 

ForsakenHarmony:
New version of #192 potentially a bit better

The error name is maybe not ideal

### DIFF
--- a/src/import.rs
+++ b/src/import.rs
@@ -162,11 +162,7 @@ pub fn import_image_data(
     Ok(images)
 }
 
-fn import_impl(path: &Path) -> Result<Import> {
-    let base = path.parent().unwrap_or(Path::new("./"));
-    let file = fs::File::open(path).map_err(Error::Io)?;
-    let reader = io::BufReader::new(file);
-    let Gltf { document, blob } = Gltf::from_reader(reader)?;
+fn import_impl(Gltf { document, blob }: Gltf, base: &Path) -> Result<Import> {
     let buffer_data = import_buffer_data(&document, base, blob)?;
     let image_data = import_image_data(&document, base, &buffer_data)?;
     let import = (document, buffer_data, image_data);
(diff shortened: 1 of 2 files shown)


src/import.rs
```diff
@@ -203,5 +201,32 @@ fn import_impl(path: &Path) -> Result<Import> {
 pub fn import<P>(path: P) -> Result<Import>
     where P: AsRef<Path>
 {
-    import_impl(path.as_ref())
+    let path = path.as_ref();
+    let base = path.parent().unwrap_or(Path::new("./"));
+    let file = fs::File::open(path).map_err(Error::Io)?;
+    let reader = io::BufReader::new(file);
+    import_impl(Gltf::from_reader(reader)?, Some(base))
```
alteous: We need to keep this similiar to before or else the compiler will generate lots of code, once for every `P` type. I suggest we add another intermediate function:

```rust
// New function (name could be improved!)
fn import_path(path: &Path) -> Result<Import> {
    ...
    import_impl(Gltf::from_reader(reader)?, Some(base))
}

// As you have
fn import_impl(Gltf { document, blob }: Gltf, base: Option<&Path>) -> Result<Import> {
    ...
}

// As before
pub fn import<P>(path: P) -> Result<Import>
    where P: AsRef<Path>,
{
    import_path(path.as_ref())
}
```

We could also make the slice parameter more generic:
```rust
fn import_slice_impl(slice: &[u8]) -> Result<Import> {
    import_impl(Gltf::from_slice(slice)?, None)
}

pub fn import_slice<S>(slice: S) -> Result<Import>
    where S: AsRef<[u8]>,
{
    import_slice_impl(slice.as_ref())
}
```

src/lib.rs
```diff
@@ -107,7 +107,7 @@ pub mod image;
 
 /// The reference importer.
 #[cfg(feature = "import_data_reference")]
-mod import;
+pub mod import;
```
alteous: Could we avoid making this module public?

src/lib.rs
```diff
@@ -207,6 +210,10 @@ pub enum Error {
     #[cfg(feature = "import")]
     MissingBlob,
 
+    /// An external file is referenced in a slice only import without path
+    #[cfg(feature = "import")]
+    ExternalReferenceInSliceImport,
```
alteous: This is a reasonable approach.

src/lib.rs
ForsakenHarmony: Yeah, not sure why I had that in there

src/import.rs
ForsakenHarmony: Done I think 👌

ForsakenHarmony:
🎉

Issue #259, opened 2019-10-12
CubicSpline has same documentation as CatmullRomSpline


rukai:
Looks like the CubicSpline documentation was copy pasted from the CatmullRomSpline documentation and then not corrected.
https://docs.rs/gltf/0.14.0/gltf/animation/enum.Interpolation.html

I looked up in the specs to see what it was supposed to be, but it doesn't exist??? So I'm a little confused, if it should actually exist at all.
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0

alteous:
Oops, well spotted! Catmull-Rom splines were removed by https://github.com/KhronosGroup/glTF/pull/1171 back in December 2017. I kept it around in case somebody was depending on it but now that seems an unlikely scenario. I suggest we remove it entirely.

PR #260, opened 2019-10-12
Fix cubic spline docs

rukai:
I realized the docs here are just copied from the gltf spec, so I made a PR for this myself.

Probably closes https://github.com/gltf-rs/gltf/issues/259

Although I'm still not sure why the catmull rom spline is not in the gltf specs.


alteous:
Thanks for the PR! I suggest we remove the `CatmullRomSpline` variant too.

PR #263, opened 2019-10-17
Add support for KHR_materials_unlit

adrien-ben:
Hi there,

I'd like to implement support for `KHR_materials_unlit` in my gltf viewer. Here is a PR adding support for it in gltf.


src/material.rs
```diff
@@ -157,6 +157,17 @@ impl<'a> Material<'a> {
         self.json.emissive_factor.0
     }
 
+    /// Specifies whether the material is unlit.
+    /// 
+    /// * When this value is false, primitive should be renderer using Physically-Based Rendering (PBR) methodology.
+    /// * When this value is true, primitive should not be renderer using Physically-Based Rendering (PBR) methodology.
+    #[cfg(feature = "KHR_materials_unlit")]
+    pub fn is_unlit(&self) -> bool {
```
alteous: Could we call this `unlit` instead?

```rust
pub fn unlit(&self) -> bool { ... }
```

src/material.rs
```diff
@@ -157,6 +157,17 @@ impl<'a> Material<'a> {
         self.json.emissive_factor.0
     }
 
+    /// Specifies whether the material is unlit.
+    /// 
+    /// * When this value is false, primitive should be renderer using Physically-Based Rendering (PBR) methodology.
+    /// * When this value is true, primitive should not be renderer using Physically-Based Rendering (PBR) methodology.
```
alteous: The `KHR_materials_unlit` is a suggestion to not use the 'core' PBR values. https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_unlit#overview

I suggest we reword something like:

```
Returns `true` if the [`KHR_materials_unlit`] property was specified, in which case the renderer should prefer to ignore all PBR values except `baseColor`.

[`KHR_materials_unlit`](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_unlit#overview)
```

src/material.rs
adrien-ben: Absolutly!

src/material.rs
adrien-ben: You're right, it's just a suggestion, I changed it. Thanks.

adrien-ben:
Thanks :)

PR #264, opened 2019-10-27
Relax proxy lifetimes

Ralith:
Previously, every accessor method borrowed directly from its object, which makes certain traversals difficult. Instead, we can thread through a single same lifetime representing the root Gltf object.

Only updated `Scene` for now; this should probably be done pervasively, but I didn't want to put in the grunt work without buy-in.


alteous:
Thanks for the PR! As you point out, this partly resolves the long standing lifetime issue.

PR #267, opened 2019-11-02
Relax additional lifetimes

Ralith:
Followup for #264. Tried to get everything, but good odds I missed a few corners.

Aside all the low hanging fruit, I made a couple more significant changes this time: the `Primitives` iterator and the `Primitive` struct itself now capture `Mesh` by value. This increases their size by two pointers in exchange for avoiding narrowing of lifetimes. An alternative approach would be to add a distrinct lifetime parameter for the `Mesh` references alone, but that would be a breaking change, and is less flexible. Considering how trivial these copies should be for LLVM to optimize, I think the chosen solution is ideal.

### DIFF
--- a/src/accessor/mod.rs
+++ b/src/accessor/mod.rs
@@ -130,7 +130,7 @@ impl<'a> Accessor<'a> {
     }
 
     /// Optional application specific data.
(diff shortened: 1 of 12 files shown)


alteous:
Thanks! :+1:

PR #269, opened 2019-11-12
Make buffer_views in sparse accessors optional

### LINKED ISSUE #266: buffer_view should be optional in sparse accessors
It seems like it could be useful to be able to omit the initial values in sparse accessors:

> A sparse accessor differs from a regular one in that bufferView property isn't required. When it's omitted, the sparse accessor is initialized as an array of zeros of size (size of the accessor element) * (accessor.count) bytes.
(taken from https://github.com/KhronosGroup/glTF/tree/master/specification/2.0)

I'm not sure how big of a change it would be to make buffer_view an Option<Index> though.

elrnv:
Fixes #266

For those who use blender, current stable blender build actually doesn't currently support importing sparse accessors without buffer_views!

Reproducing the error message below:
```
Traceback (most recent call last):
  File "/Applications/Blender.app/Contents/Resources/2.80/scripts/addons/io_scene_gltf2/__init__.py", line 528, in execute
    return self.import_gltf2(context)
  File "/Applications/Blender.app/Contents/Resources/2.80/scripts/addons/io_scene_gltf2/__init__.py", line 549, in import_gltf2
    BlenderGlTF.create(self.gltf_importer)
  File "/Applications/Blender.app/Contents/Resources/2.80/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_gltf.py", line 41, in create
    BlenderScene.create(gltf, scene_idx)
  File "/Applications/Blender.app/Contents/Resources/2.80/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_scene.py", line 102, in create
    BlenderAnimation.anim(gltf, anim_idx, node_idx)
  File "/Applications/Blender.app/Contents/Resources/2.80/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation.py", line 30, in anim
    BlenderNodeAnim.anim(gltf, anim_idx, node_idx)
  File "/Applications/Blender.app/Contents/Resources/2.80/scripts/addons/io_scene_gltf2/blender/imp/gltf2_blender_animation_node.py", line 104, in anim
    values = BinaryData.get_data_from_accessor(gltf, animation.samplers[channel.sampler].output)
  File "/Applications/Blender.app/Contents/Resources/2.80/scripts/addons/io_scene_gltf2/io/imp/gltf2_io_binary.py", line 52, in get_data_from_accessor
    bufferView = gltf.data.buffer_views[accessor.buffer_view]  # TODO initialize with 0 when not present!
TypeError: list indices must be integers or slices, not NoneType

location: <unknown location>:-1
```


src/accessor/mod.rs
```diff
@@ -109,8 +109,10 @@ impl<'a> Accessor<'a> {
     }
 
     /// Returns the buffer view this accessor reads from.
-    pub fn view(&self) -> buffer::View<'a> {
-        self.document.views().nth(self.json.buffer_view.value()).unwrap()
+    ///
+    /// This may be `None` if the corresponding accessor is sparse.
+    pub fn view(&self) -> Option<buffer::View<'a>> {
+        self.json.buffer_view.map(|view| self.document.views().nth(view.value()).unwrap())
```
alteous: We need to add a validation check for this otherwise the unwrap might fail. A new `Validate` implementation will be required for `gltf_json::Accessor` to replace the [automatically derived one](https://github.com/gltf-rs/gltf/blob/36fbd771acad62d7f8a633fec32ebdefdf6f3d4a/gltf-json/src/accessor.rs#L191).

elrnv: That unwrap is for the nth call for indexing buffer views (I think). It was there before this PR. Do you mean that we need a validation for checking that buffer_view can only be omitted if sparse is present? Anyway I added a tentative custom Validate impl for your review.

alteous: Yes, that's what I meant. This looks good, thanks.

PR #246, opened 2019-08-19
Add sparse accessor iterators

alteous:
Fixes #154

PR #271, opened 2019-11-17
Implement ExactSizeIterator for accessor::util::Iter

aloucks:
This trait was implemented for `Iter` prior to 0.14. It appears that
the missing trait implementation was an oversight when sparse
accessors were implemented in #246.

### DIFF
--- a/src/accessor/util.rs
+++ b/src/accessor/util.rs
@@ -35,6 +35,15 @@ impl<'a, T: Item> Iterator for Iter<'a, T> {
     }
 }
 
+impl<'a, T: Item> ExactSizeIterator for Iter<'a, T> {
+    fn len(&self) -> usize {
+        match self {
+            &Iter::Standard(ref iter) => iter.len(),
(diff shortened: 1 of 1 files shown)


alteous:
Thanks for the PR! It's not an oversight per se, I didn't think it made sense at the time. After re-reading the [`ExactSizeIterator` docs](https://doc.rust-lang.org/stable/std/iter/trait.ExactSizeIterator.html) I don't think this implementation is appropriate.

I think we would need to do something like this instead:

```rust
impl<'a, T: Item> ExactSizeIterator for SparseIter<'a, T> {}
impl<'a, T: Item> Iterator for SparseIter<'a, T> {
    ...
    fn size_hint(&self) -> usize {
        let hint = self.values.len() - self.counter;
        (hint, Some(hint))
    }
}
```

aloucks:
@alteous I've updated this based on your feedback and also added the other methods to `Iter` that were implemented for `ItemIter`.

alteous:
LGTM, thanks. :+1:

PR #275, opened 2020-02-08
Try guessing image format when MIME type is missing or invalid

dariost:
When the MIME type is missing gltf can try guessing the file format by using `image::guess_format`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ KHR_lights_punctual = ["gltf-json/KHR_lights_punctual"]
 KHR_materials_pbrSpecularGlossiness = ["gltf-json/KHR_materials_pbrSpecularGlossiness"]
 KHR_materials_unlit = ["gltf-json/KHR_materials_unlit"]
 image_jpeg_rayon = ["image/jpeg_rayon"]
+guess_mime_type = []
 
 [[example]]
 name = "gltf-display"
--- a/src/import.rs
+++ b/src/import.rs
@@ -112,12 +112,17 @@ pub fn import_image_data(
     buffer_data: &[buffer::Data],
 ) -> Result<Vec<image::Data>> {
     let mut images = Vec::new();
+    #[cfg(feature = "guess_mime_type")]
     let guess_format = |encoded_image: &[u8]| {
         match image_crate::guess_format(encoded_image) {
             Ok(image_crate::ImageFormat::PNG) => Some(Png),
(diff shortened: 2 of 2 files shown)


alteous:
Hi, thanks for your pull request. I'm curious what problem this is going to solve. When unspecified, the media type and encoding of a data URI is assumed to be plaintext and ASCII respectively ([RFC 2397](https://tools.ietf.org/html/rfc2397)).

If an embedded image in a glTF file does not specify both the media type and `base64` tag then it is either non-standard glTF or a malformed data URI, though [the spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#images) also says:

> **Implementation Note:** When image data is provided by uri and mimeType is defined, client implementations should prefer JSON-defined MIME Type over one provided by transport layer.

I notice we don't handle this particular case. Would your data be imported correctly if we did?

I'm not overly keen on accepting malformed data because over time users may depend on it.

dariost:
Hi, let me start by saying that in principle I agree with you 100% on not accepting malformed data to not cause bad habits, but I thought that this could be useful to other users in my situation.

When exporting in `.glb` format from 3ds Max using the [Khronos' recommended plugin](https://github.com/KhronosGroup/glTF#gltf-tools) (the one based on BabylonJS) the MIME type is an empty string, causing this crate to raise an `UnsupportedImageEncoding` error.

This is, of course, a bug with the exporter, but I have to deal with a lot of `.glb` files that were exported from it, and were the original project file has been lost, so I figured it would be easier to make the image decoding process more invalid-MIME-type agnostic.

alteous:
It's not quite an empty string but rather the second half of the MIME type is missing:
```json
"images":[{"mimeType":"image/","bufferView":3}]
```

It failed [validation](https://github.khronos.org/glTF-Validator) with the message:

```json
{
    "code": "VALUE_NOT_IN_LIST",
    "message": "Invalid value 'image/'. Valid values are ('image/jpeg', 'image/png').",
    "severity": 1,
    "pointer": "/images/0/mimeType"
},      
```

I wonder whether there is a default subtype for this case. A quick web search suggests no. As a temporary measure, you could find/replace `image/` with `image/png` in the binaries.

I'd feel more comfortable accepting non-standard image types if we were to make doing do an opt-in feature, e.g. `features = ["allow-non-standard-image-formats"]` or words to that effect.

dariost:
I've added a `guess_mime_type` feature as you suggested.

alteous:
Looks good to me, thanks. :+1:

PR #280, opened 2020-03-15
Have docs.rs document all features

### LINKED ISSUE #233: Documentation shows only default features
The documentation automatically hosted on [docs.rs](https://docs.rs/gltf/) exposes only the default features. We should figure out how to enable them all or else consider hosting the documentation elsewhere.

aloucks:
Fixes #233

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,3 +60,6 @@ path = "examples/roundtrip/main.rs"
 [[example]]
 name = "gltf-tree"
 path = "examples/tree/main.rs"
+
+[package.metadata.docs.rs]
+all-features = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,3 +63,4 @@ path = "examples/tree/main.rs"
 
 [package.metadata.docs.rs]
 all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
--- a/src/accessor/mod.rs
+++ b/src/accessor/mod.rs
@@ -62,6 +62,7 @@ pub use json::accessor::Type as Dimensions;
 
 /// Utility functions.
 #[cfg(feature = "utils")]
+#[cfg_attr(docsrs, doc(cfg(feature = "utils")))]
 pub mod util;
 
 /// Contains data structures for sparse storage.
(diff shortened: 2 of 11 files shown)


alteous:
Thanks for the PR! I've been hesitant to implement this for a while because I think there is more to do than merely expose all the features. I think we ought to inform the user somehow about what features belong to each feature flag, similar to the [`syn` crate](https://docs.rs/syn/1.0.16/syn/enum.Item.html) for example, otherwise the documentation could be confusing.

aloucks:
There's an unstable `doc_cfg` attribute that can do mostly what you're looking for. It has some quirks when you re-export from modules, but is otherwise pretty nice.

It can also be enabled for `docs.rs` with config flags. Here's an example of when I added it to the  [hex](https://docs.rs/hex/0.4.2/hex/) crate with the serde feature: https://github.com/KokaKiwi/rust-hex/pull/38/files

alteous:
That looks great, exactly what I had in mind. What does it mean for a `doc_cfg` attribute to be unstable? Would it require a nightly compiler?

aloucks:
It does require a nightly compiler which is why it needs some special flags set:

```toml
[package.metadata.docs.rs]
all-features = true
rustdoc-args = ["--cfg", "docsrs"] 
```

lib.rs:
```rust
#![cfg_attr(docsrs, feature(doc_cfg))]
```

toggling it:
```rust
#[cfg(feature = "my_feature")]
#[cfg_attr(docsrs, doc(cfg(feature = "my_feature")))]
pub mod serde;
```

Basically, you use `--cfg docsrs` to turn the feature on whenever it's run from `docs.rs` (which uses a nightly compiler). You can test it locally with `RUSTDOCFLAGS="--cfg docsrs" cargo doc`

alteous:
I gave it a try. The output looks nice but there are missing labels for re-exports as you mentioned. Is there a workaround?

aloucks:
I haven't found a workaround.

alteous:
Is there any progress here? I'd like to add my own ad-hoc labels in the meantime.

aloucks:
I think if you make the modules public it will work as expected.

aloucks:
@alteous I've added the tags everywhere.

    RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --open --all-features

alteous:
The Travis run isn't showing up which is weird but otherwise seems to be passing: https://travis-ci.org/github/gltf-rs/gltf/builds/666617711?utm_medium=notification&utm_source=github_status

alteous:
Marvellous! Thank you very much. :smile: 

I will publish the changes by the end of the week.

PR #282, opened 2020-03-28
Implement `nth` for the primary iterators

aloucks:
The default implementation of `std::iter::Iterator::nth` effectively just calls `next()` for `0..n`. 

```rust
#[stable(feature = "rust1", since = "1.0.0")]
fn nth(&mut self, mut n: usize) -> Option<Self::Item> {
    for x in self {
        if n == 0 {
            return Some(x);
        }
        n -= 1;
    }
    None
}
```

The iterators in this crate are adapters over `std::slice::Iter`, which has a specialized version of `nth` to make it `O(1)`. By implementing `nth` and calling the inner iterator method we can effectively turn `nth` into an indexed array access instead of looping through all of the items. 

https://doc.rust-lang.org/src/core/slice/mod.rs.html#3211-3306
```rust
fn nth(&mut self, n: usize) -> Option<$elem> {
    if n >= len!(self) {
        // This iterator is now empty.
        if mem::size_of::<T>() == 0 {
            // We have to do it this way as `ptr` may never be 0, but `end`
            // could be (due to wrapping).
            self.end = self.ptr.as_ptr();
        } else {
            unsafe {
                // End can't be 0 if T isn't ZST because ptr isn't 0 and end >= ptr
                self.ptr = NonNull::new_unchecked(self.end as *mut T);
            }
        }
        return None;
    }
    // We are in bounds. `post_inc_start` does the right thing even for ZSTs.
    unsafe {
        self.post_inc_start(n as isize);
        Some(next_unchecked!(self))
    }
}
```

https://doc.rust-lang.org/src/core/slice/mod.rs.html#3167
```rust
// Helper function for moving the start of the iterator forwards by `offset` elements,
// returning the old start.
// Unsafe because the offset must not exceed `self.len()`.
#[inline(always)]
unsafe fn post_inc_start(&mut self, offset: isize) -> * $raw_mut T {
    if mem::size_of::<T>() == 0 {
        zst_shrink!(self, offset);
        self.ptr.as_ptr()
    } else {
        let old = self.ptr.as_ptr();
        self.ptr = NonNull::new_unchecked(self.ptr.as_ptr().offset(offset));
        old
    }
}
```

### DIFF
--- a/src/animation/iter.rs
+++ b/src/animation/iter.rs
@@ -27,11 +27,37 @@ impl<'a> Iterator for Channels<'a> {
     fn next(&mut self) -> Option<Self::Item> {
         self.iter.next().map(|json| Channel::new(self.anim.clone(), json))
     }
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.iter.size_hint()
+    }
+    fn count(self) -> usize {
+        self.iter.count()
+    }
+    fn last(self) -> Option<Self::Item> {
+        let anim = self.anim;
+        self.iter.last().map(|json| Channel::new(anim, json))
+    }
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        self.iter.nth(n).map(|json| Channel::new(self.anim.clone(), json))
+    }
 }
 
 impl<'a> Iterator for Samplers<'a> {
     type Item = Sampler<'a>;
     fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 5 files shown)


alteous:
This is a maintenance pain point. We should try to find a way to avoid this code duplication in future.

alteous:
FYI, I'll publish a new release today.

Jasper-Bekkers:
Thanks, as shown in the linked issue on large files this really adds up. I measured a ~1000x speedup in a 1.5G scene file.

PR #283, opened 2020-03-28
Fix deprecation warnings with `std::error::Error::description`.

aloucks:
Fixes these warnings:
```
warning: use of deprecated item 'std::error::Error::description': use the Display impl or to_string()
   --> src\lib.rs:560:35
    |
560 |             Error::Io(ref e) => e.description(),
```


alteous:
Great, thanks!

PR #288, opened 2020-05-05
Support URIs with embedded data when importing from a slice

aclysma:
I have data in this form:

```
    "buffers" : [
        {
            "byteLength" : 160244,
            "uri" : "data:application/octet-stream;base64,wsVHPgAAgD++FHu/ws...
```

Currently if trying to load this data using `gltf::import_slice(&bytes)` this will fail because as the code is written, it assumes a non-None path

I think it would be more accurate to move this check into the read() function allowing for the base64 case to work with a None path.

### DIFF
--- a/src/import.rs
+++ b/src/import.rs
@@ -53,12 +53,15 @@ impl<'a> Scheme<'a> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


alteous:
Hi, thanks for your PR and sorry for the delay. This seems very sensible. :+1:

PR #293, opened 2020-09-13
Add support for KHR_materials_transmission and KHR_materials_ior

eric1221bday:
The `KHR_mterials_transmission` extension can be found [here](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_materials_transmission/README.md) and is more or less finalized.

The `KHR_materials_ior` extension can be found [here](https://github.com/DassaultSystemes-Technology/glTF/tree/KHR_materials_ior/extensions/2.0/Khronos/KHR_materials_ior). It's not yet finalized but at this point it's very unlikely the interface itself would change, and it looks like all that's left is for them to better document and come up with examples.


alteous:
Thanks for your PR! Looks good to me.

PR #299, opened 2021-02-28
Make alpha cutoff optional.

FrankenApps:
I don't know if this is a welcome change, but I think it should be possible to create _materials_ without an `alphaCutoff` field. 
The reason is that when `"alphaMode" : "BLEND"` is used, this field will be ignored (see this [Issue](https://github.com/KhronosGroup/glTF/issues/1158)).

Therefore the Khronos **glTF-Validator** shows this warning, when the `alphaCutoff` field is present and the `alphaMode` is `BLEND`:
validation_warning

There are also official [examples](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AlphaBlendModeTest) where this field is ommited for `"alphaMode" : "BLEND"`.

Apart from that I changed the Documentation comment for `"BLEND"` to the following, because it seems it was previously just copied from `"MASK"`:

> The alpha value is used, to determine the transparency of the rendered output.
    The alpha cutoff value is ignored.

### DIFF
--- a/src/material.rs
+++ b/src/material.rs
@@ -51,12 +51,7 @@ impl<'a> Material<'a> {
 
     ///  The optional alpha cutoff value of the material.
     pub fn alpha_cutoff(&self) -> Option<f32> {
-        if self.json.alpha_cutoff.is_some() {
-            return Some(self.json.alpha_cutoff.unwrap().0);
-        }
-        else{
-            None
-        }
+        self.json.alpha_cutoff.map(|value| value.0)
     }
 
     /// The alpha rendering mode of the material.  The material's alpha rendering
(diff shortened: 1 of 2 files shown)


src/material.rs
```diff
@@ -49,9 +49,14 @@ impl<'a> Material<'a> {
         self.index
     }
 
-    ///  The alpha cutoff value of the material.
-    pub fn alpha_cutoff(&self) -> f32 {
-        self.json.alpha_cutoff.0
+    ///  The optional alpha cutoff value of the material.
+    pub fn alpha_cutoff(&self) -> Option<f32> {
+        if self.json.alpha_cutoff.is_some() {
```
aloucks, line 54: I think you could simplify this as:

```rust
pub fn alpha_cutoff(&self) -> Option<f32> {
    self.json.alpha_cutoff.map(|value| value.0)
}
```

FrankenApps: Done.

PR #304, opened 2021-06-22
added cesium rtc

bkuster:
# The Issue
The loader does not support the [CESIUM_RTC](https://github.com/KhronosGroup/glTF/blob/master/extensions/1.0/Vendor/CESIUM_RTC/README.md) extension.

# Solution
- Added the `CesiumRtc` struct with one filed `center: [f64; 3]`.
- Added Validate for `[f64; 3]`.
- The `extension::Root` now has a field `cesium_rtc: Option<CesiumRtc>`.

# Issues
Im still unsure of what to do with the MODELVIEW semantic (apparenty the extension [is supported by gltf 2.0](https://github.com/KhronosGroup/glTF/issues/1159), but they never adapted the spec), so I opted to simply ignore it. Not quite sure what happened to the 1.0 techniques and parameters anyway :shrug:.


IcanDivideBy0:
Hello, this crate explicitly states it's for glTF 2.0. As far as I understand, CESIUM_RTC is a glTF 1.0 extension, so i'm not sure this should be added. @aloucks any thought on this ?

bkuster:
So, this is fairly stale and I guess we will never get the MODELVIEW in here. Since B3DM now provides a batch table semantic RTC_CENTER (see [here](https://docs.opengeospatial.org/cs/18-053r2/18-053r2.html#10.1.4)) I think we can close this.

PR #305, opened 2021-08-01
Stop computing the glb length in the example

wlinna:
`to_writer` [computes the length internally](https://github.com/gltf-rs/gltf/blob/master/src/binary.rs#L174), and the length provided in the header is ignored. Thus computing it yourself only adds complexity.

### DIFF
--- a/examples/export/main.rs
+++ b/examples/export/main.rs
@@ -22,10 +22,6 @@ struct Vertex {
     color: [f32; 3],
 }
 
-fn align_to_multiple_of_four(n: &mut u32) {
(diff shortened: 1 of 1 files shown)


IcanDivideBy0:
Hello, thanks!
It feels a bit weird to have a random zero in the middle of an example. IMO we should either add a comment to explain that length is recomputed anyways thus we can set the header length to whatever, or add an `impl Default for Header`. There's no much room for changing anything in the header:
* magic: Must be b"glTF".
* version: Must be 2.
* length: is ignored most of the time ?

PR #306, opened 2021-08-18
Fix `Normalize` for `u8`

h3r2tic:
Hey! This fixes what looks like copy-paste errors in `Normalize` for `u8`

* The `impl Normalize<f32> for u8` was causing `u8` vertex colors to appear black, and now exhibits the correct behavior.
* The `impl Normalize<u16> for u8` I did not test, but it follows the patterns of the other integer `Normalize` implementations.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -614,11 +614,11 @@ impl Normalize<i16> for u8 {
 }
 
 impl Normalize<u16> for u8 {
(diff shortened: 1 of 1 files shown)


IcanDivideBy0 (approved):

alteous:
Oops, well spotted. We'll mark this is a bug in the release notes.

PR #312, opened 2021-11-02
follow spec to change more fields to optional

wsw0108:
Follow the spec https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html

### DIFF
--- a/gltf-json/src/animation.rs
+++ b/gltf-json/src/animation.rs
@@ -127,7 +127,8 @@ pub struct Target {
     pub extras: Extras,
     
     /// The index of the node to target.
-    pub node: Index<scene::Node>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub node: Option<Index<scene::Node>>,
     
     /// The name of the node's property to modify or the 'weights' of the
     /// morph targets it instantiates.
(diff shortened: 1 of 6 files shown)


aloucks:
I have mixed feelings about this. The current behavior will fill the attribute with the default value when it's not present. This PR seems like more of an optimization for serialization. Perhaps an alternate solution would be to skip serialization when the value is equal to the default value?

It's going to be a lot more cumbersome to have to consult the spec to discover the default values with this change. One way to mitigate this would be to provide public constants with the defaults. 

e.g.

```rust
let double_sided = material.double_sided.unwrap_or(Material::DEFAULT_DOUBLE_SIDED);
```

But this begs the question: What should something like `glfs_json::Material::default()` produce in this case? Should the default for `double_sided` be `None` or  `Some(false)`?

I think the most well-rounded solution would be to skip serialization when the value equals the spec-default, and only change things to `Option<T>` if/when there is no default value.

IcanDivideBy0:
I agree with @aloucks on this, as a user of this crate, i don't want to spend my time browsing the gltf spec to find what is the default value for each property. The actual behavior seems fine to me, so aside from saving a few bytes from the exported gltf, i don't see much value here.
Otherwise, having the default value accessible as a struct constant seems reasonable, it also makes loading your models cumbersome where it doesn't need to.

wsw0108:
Thanks for your comments, then I will close it.

PR #315, opened 2021-11-14
Convert R8G8B8 textures to R8G8B8A8 when importing for better compatibility with GPUs

expenses:
If you search for `R8G8B8_` on either https://vulkan.gpuinfo.org/listlineartilingformats.php or https://vulkan.gpuinfo.org/listoptimaltilingformats.php you'll see that few GPUs are capable of using them in any meaningful way:

compatibility

No Nvidia or AMD GPUs can sample them, for example: https://vulkan.gpuinfo.org/listdevicescoverage.php?lineartilingformat=R8G8B8_SRGB&featureflagbit=SAMPLED_IMAGE&platform=all.

This is a problem when importing glTF files, as the referenced images often do not have alpha channels. The solution I've come up with here is to remove the `R8G8B8` format and convert `DynamicImage::ImageRgb8`s to RGBA instead.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -15,9 +15,6 @@ pub enum Format {
     /// Red, green.
     R8G8,
 
-    /// Red, green, blue.
-    R8G8B8,
-
     /// Red, green, blue, alpha.
     R8G8B8A8,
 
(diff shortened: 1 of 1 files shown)


FrankenApps:
I don't see a reason why the library itself should do this. The user can always convert his image to rgba and then use `DynamicImage::ImageRgba8(_)` if he wants the better GPU support. 
This library should instead adhere to the gltf spec.

expenses:
Well the main reason for having this in the crate is that someone would need to convert the loaded `gltf::image::Data` structs into `image::RgbImage`s and then convert them to RGBA. It'd be a lot nicer to convert them when they're `image::DynamicImage`s in the crate. Even if this isn't the default functionality, it would be nice to have it as a helper function.

FrankenApps:
I see. I guess a helper function would make sense then, but I personally would not like the library to do this by default.

expenses:
I can't find a good way to do this nicely in the crate, but it's not too bad to do it the way I mention above, so I'm going to close this for now.

PR #333, opened 2022-02-19
Update image dependency from 0.23 to 0.24

kirdaybov:
When building our repository that depends on `gltf` we found out that it is not using the last `image` crate. We would appreciate if the dependency could be brought up to date.


src/image.rs
```diff
@@ -145,15 +145,16 @@ impl Data {
...
             DynamicImage::ImageRgba8(_) => Format::R8G8B8A8,
-            DynamicImage::ImageBgr8(_) => Format::B8G8R8,
-            DynamicImage::ImageBgra8(_) => Format::B8G8R8A8,
             DynamicImage::ImageLuma16(_) => Format::R16,
             DynamicImage::ImageLumaA16(_) => Format::R16G16,
             DynamicImage::ImageRgb16(_) => Format::R16G16B16,
             DynamicImage::ImageRgba16(_) => Format::R16G16B16A16,
+            DynamicImage::ImageRgb32F(_) => Format::R32G32B32FLOAT,
+            DynamicImage::ImageRgba32F(_) => Format::R32G32B32A32FLOAT,
+            _ => unreachable!("Unmatched DynamicImage format"),
```
kirdaybov, line 154: `enum DynamicImage` is `#[non_exhaustive]` since 0.24

IcanDivideBy0:
Thank you!
We should avoid panics when possible, i've added proper error handling in #344.
Since this is private API we don't need a major release to handle the change

src/image.rs
```diff
@@ -145,15 +145,16 @@ impl Data {
             DynamicImage::ImageLumaA8(_) => Format::R8G8,
             DynamicImage::ImageRgb8(_) => Format::R8G8B8,
             DynamicImage::ImageRgba8(_) => Format::R8G8B8A8,
-            DynamicImage::ImageBgr8(_) => Format::B8G8R8,
-            DynamicImage::ImageBgra8(_) => Format::B8G8R8A8,
```
alteous, line 149: Why have these cases been removed?

kirdaybov: They have been removed from the `DynamicImage` of the `image` crate. You can take a look at the corresponding discussions https://github.com/image-rs/image/pull/1482 https://github.com/image-rs/image/issues/1671

PR #334, opened 2022-02-21
Avoid panics on image import

IcanDivideBy0:
follow up #333


src/lib.rs
```diff
@@ -567,6 +572,10 @@ impl std::fmt::Display for Error {
             #[cfg(feature = "import")]
             Error::UnsupportedImageEncoding => write!(f, "unsupported image encoding"),
             #[cfg(feature = "import")]
+            Error::UnsupportedImageFormat(image) => {
+                write!(f, "unsupported image encoding: {:?}", image)
```
aloucks, line 576: The debug representation of `DynamicImage` contains the entire byte array. I doubt that's what you intended here. 

This would be a better alternative:

```rust
write!(f, "unsupported image encoding: {:?}", image.color())
```

IcanDivideBy0: thanks! i changed it in #335

Issue #336, opened 2022-03-03
Determinism issue when serializing Primitive::attributes


kirdaybov:
When serializing Gltf::document with gltf::json::serialize::to_vec we get different ordering on the `attributes` map of a primitive. This happens due to the dependence of the behavior of the hash function on the starting condition of the program. As a result, we get a nondeterministic result:
`"attributes":{"NORMAL":1,"TEXCOORD_0":2,"POSITION":0}` vs `"attributes":{"POSITION":0,"NORMAL":1,"TEXCOORD_0":2}`

kirdaybov:
Proposed fix: https://github.com/gltf-rs/gltf/pull/337

alteous:
Why is this an issue? A hash map is likely to have better memory locality than a binary tree.

kirdaybov:
In my case I wanted to serialize the import result in our data pipeline (document + buffers + images) as a single blob. The reason is to have an almost raw data without the need for a virtual file system (since GLTF can have external links to textures or .bin file). The data pipeline expects data serialization to be deterministic.

The only thing that behaves non deterministically during serialization is the attributes. Since the amount of data stored in the attribute map is small (<10 I believe) the difference of impact on memory and performance between HashMap and BTreeMap shouldn't be significant.

kirdaybov:
Closed with https://github.com/gltf-rs/gltf/pull/337

PR #337, opened 2022-03-03
Fix Primitive::attributes serialization determinism issue

kirdaybov:
HashMap uses a hash function that behaves differently based on the moment of the initialization of the program. That leads to the different iteration orders during serialization. Replacing HashMap with BTreeMap fixes the issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The `gltf` crate adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Fix relative file path imports with url encoded characters.
 - Update dependency on `image` crate from 0.23 to 0.24.
+- Fix serialization determinism related to using HashMap for Primitive::attributes. Now uses BTreeMap instead.
 
 ## [1.0.0] - 2022-01-29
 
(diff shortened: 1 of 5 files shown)


kirdaybov:
#336

alteous:
Hi, thanks for your PR. I have the exact same problem but I'm replacing `HashMap` with `BTreeMap` might affect performance. Perhaps we could implement this as a feature option instead, something along the lines of `deterministic_serialization`.

kpreid:
> I'm concerned replacing `HashMap` with `BTreeMap` might affect performance.

The major cost of a `BTreeMap` vs. `HashMap` is that it requires more pointer indirection to follow the tree nodes. But the value of the “B” constant actually used in Rust's `BTreeMap` is 6, meaning that the tree contains only one node unless there are more than `2*B - 1 = 11` entries (attributes) stored in it — which is the same amount of indirection a `HashMap`'s storage needs too. So, this is unlikely to be a problem except for (1) very complex attribute sets (2) which are repeatedly accessed directly from the `gltf` data structure instead of iterated once and copied to the application's own data structure.

alteous:
@kpreid, thanks for your input, the pointer indirection was indeed my concern. I assumed the `B` in `BTreeMap` stood for binary without looking any further into it. I stand corrected. 😄 

It's unlikely the number of vertex attributes would exceed 11 so I see no issues merging as-is.

@kirdaybov, I apologise if my misunderstanding caused any frustration.

kirdaybov:
No hard feelings ;)

PR #339, opened 2022-04-11
Update dependencies: base64 and approx

emilk:
Latest are always best? :)

(also less likely to cause duplicates for dependent crates)


alteous:
Thanks :)

PR #340, opened 2022-04-24
Don't serialize `Extras`

### LINKED ISSUE #230: Empty extras are serialized
Example

```json
{
    "extras": {}
}
```

We don't need these to be serialized.

digama0:
...when extras feature is disabled.

Fixes #230


alteous:
Thanks!

PR #341, opened 2022-04-26
Add more public import facilities

### LINKED ISSUE #222: Import buffer data separately from other import data.
Not sure how to import buffer data alone, because the import module is private, so the function `import_buffer_data` appears inaccessible. It appears that I only have access to opening the gltf file, which doesn't import buffer data, or calling `gltf::import`, which imports both the buffer data and the image data. I already load images on my own, for various reasons, but importing buffer data is useful. Perhaps I am missing something?

digama0:
Fixes #222. New public API:

* `gltf::import` module (debatable: it would also be possible to keep this private and re-export the other functions)
* `gltf::import::import` and `gltf::import::import_slice`: These were already available as `gltf::import` and `gltf::import_slice` but now you can also see them at this path
* `gltf::import::{import_buffer_data, import_image_data}`: These already existed but they are now `pub`. This is the part that actually fixes #222.
* `gltf::buffer::Data::{from_source, from_source_and_blob}`: This allows importing just one buffer and getting errors about it. Note that `import::Scheme` is *not* made public here.
* `gltf::image::Data::from_source`: Same thing but for images. Most of the logic from `import_image_data` is now in this function.

### DIFF
--- a/src/import.rs
+++ b/src/import.rs
@@ -80,6 +80,35 @@ where
     Ok(data)
 }
 
+impl buffer::Data {
+    /// Construct a buffer data object by reading the given source.
+    /// If `base` is provided, then external filesystem references will
+    /// be resolved from this directory.
+    pub fn from_source(source: buffer::Source<'_>, base: Option<&Path>) -> Result<Self> {
+        Self::from_source_and_blob(source, base, &mut None)
+    }
+
+    /// Construct a buffer data object by reading the given source.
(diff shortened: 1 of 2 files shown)


kvark:
A bit worrying to see a PR like this, which is well written with a great description, proven to be needed (based on the issue discussion), and has been left without any answer for 9 months.

PR #347, opened 2022-06-03
Export example: calculate bounding coordinates instead of using hardcoded values

juliavdkris:
I ran into this when using this function as a reference for a project, and accidentally overlooked the hardcoded min/max values. So I thought I'd contribute this to upstream for the next person wanting to use this as a reference :)


alteous:
Thanks

Issue #231, opened 2019-07-21
Support for custom extensions


alteous:
There are two formally supported extensions by the crate: `KHR_materials_pbrSpecularGlossiness` and `KHR_lights_punctual`. As discussed in #226 there is no way to retrieve any other extension data. I believe it's now possible to provide unrecognized extensions alongside the recognized ones using a combination of [`#[serde(flatten)]`](https://serde.rs/attr-flatten.html) and [`serde::value::RawValue`](https://docs.serde.rs/serde_json/value/struct.RawValue.html).

```json
{
    "extensions": {
        "KHR_lights_punctual": {},
        "EXT_some_other_extension": {}
    }
}
```

```rust
#[derive(Deserialize, Serialize)]
pub struct Root {
    #[serde(default, serialize_only_if = "Option::is_none")]
    pub extensions: extensions::Root,
}

pub mod extensions {
    #[derive(Deserialize, Serialize)]
    pub struct Root {
        #[serde(default, rename = "KHR_lights_punctual", serialize_only_if = "Option::is_none")]
        pub khr_lights_punctual: KhrLightsPunctual,
        #[serde(default, flatten)]
        pub others: HashMap<String, RawValue>,
    }
}
```

As for retrieving the data, for now I propose a simple getter:

```rust
pub mod extensions {
    pub struct Root<'a> {
        document: &'a Document,
        json: &'a json::extensions::Root,
    }

    impl<'a> Root<'a> {
        pub fn others(&self) -> &HashMap<String, RawValue> {
            &self.others
        }
    }
}
```

zellski:
Likely naive question: If this feature is implemented as suggested above, would it be reasonably easy for a user to fish out (hypothetical example) `root.others["KHR_texture_transform"]` and, if non-empty, perform a follow-up pass of serde-deserialisation on?

Having access to the raw JSON is vastly better than not having it, but it'd be even better if I could hook up my own `KhrLightsPunctual`-like struct hierarchy.

alteous:
Yes, that's the idea. I imagine user code would look something like this:

```rust
if let Some(raw_value) = root.others().get("KHR_texture_transform") {
    #[derive(Deserialize)]
    struct KhrTextureTransform {
        pub foo: i32,
        ...
    }
    let khr_texture_transform: KhrTextureTransform = serde_json::from_str(raw_value.get())?;
}
```

3c1u:
I'd prefer that the extension can be deserialized directly into `struct`.
[...]

Issue #11, opened 2017-03-24
Extensions and Extras


alteous:
The library would be much more useful if it included the extensions and extras fields that are found on most objects in the specification. Until version `0.4.0` this was achieved by having an untyped `Map<String, Value>` structure that was taken directly from `serde_json`. This was a very lazy implementation and was not the best option in terms of user-friendliness and efficiency.

I propose we add support for extensions and extras through `Extensions`* and `Extras` traits respectively. Each trait will contain user-defined types to be (de)serialised by the library. These types could then be passed through the glTF tree structure as follows:

*Edit: See updates posted below.

```rust
// User-defined data is declared with the `Extras` trait
trait Extras {
    type Accessor: Clone + Debug + Default + Deserialize + Serialize;
    ...
}

// Data targeting official extensions is declared with the `Extensions` trait
trait Extensions {
    type Accessor: Clone + Debug + Default + Deserialize + Serialize;
    ...
}

// This is the 'entry point' for the `Extras` and `Extensions` definitions
pub fn import<P, E, X>(path: P) -> Result<Root<E, X>, ImportError>
    where P: AsRef<Path>, E: Extensions, X: Extras
{
    ...
}

// The `Extensions` and `Extras` types are passed down through the glTF tree structure
struct Root<E: Extensions, X: Extras> {
    accessors: Vec<Accessor<E, X>>,
    ...
}

// Here we resolve the user-provided extensions and extras
struct Accessor<E: Extensions, X: Extras> {
    ...
    extensions: <E as Extensions>::Accessor,
    extras: <X as Extras>::Accessor,
}
```

The implementation might be a bit bonkers, but it does allow for user-data (de)serialisation to be a zero-cost abstraction.

As far as extensions are concerned, it will be the responsibility of the library to provide the structures necessary to represent official extensions. These structures will be defined in an `extensions` module. The user should not be able to implement their own `Extensions`; only those offered by the library will be supported. If an extension is required but not supported, the library will return an appropriate error during importing / exporting.

alteous:
`Extras` has been implemented as part of the [`extensions`](https://github.com/alteous/gltf/tree/extensions) branch ([example](https://github.com/alteous/gltf/blob/extensions/src/bin/app.rs)).

alteous:
### Update

[...]

PR #348, opened 2022-06-09
Add an trait for custom extensions

expenses:
This is my attempt at solving https://github.com/gltf-rs/gltf/issues/231. We add a trait that allows you to specify all the custom fields within different parts of the gltf file. Currently I've only added support for extensions on the root node because that's what I need for the [VRM](https://github.com/vrm-c/vrm-specification) spec.

The changes to the main gltf crate are pretty massive but it's mostly just adding a generic to everything. I'm pretty happy with the gltf-json changes I've made. I had to implement an `IndexEnum` enum because the validation code for indices got very awkward otherwise:

```rust
impl<T: Validate> Validate for Index<T>
where
    Root: Get<T>,
{
     fn validate<P, R, E: crate::ThirdPartyExtensions>(&self, root: &Root<E>, path: P, report: &mut R)
    where
        P: Fn() -> Path,
        R: FnMut(&dyn Fn() -> Path, validation::Error),
    {
        .
    }
}
```

would have to become

```rust
impl<T: Validate, E: crate::CustomExtensions> Validate for Index<T>
where
    Root<E>: Get<T>,
{
     fn validate<P, R, OtherE: crate::ThirdPartyExtensions>(&self, root: &Root<E>, path: P, report: &mut R)
    where
        P: Fn() -> Path,
        R: FnMut(&dyn Fn() -> Path, validation::Error),
        // Uh oh
        E = OtherE
    {
        .
    }
}
```

which I don't think is really possible to get working.

This is a breaking change as function's don't support default generics. I've tried to add default generics in a few places to reduce breakages though.

Here's what adding support for the VRM extension might look like (I'm making a crate for this):

```rust
#[derive(serde::Deserialize, serde::Serialize, Debug, gltf::derive::Validate, Clone)]
struct Vrm {
    #[serde(rename = "exporterVersion")]
    ...
}

#[derive(serde::Deserialize, serde::Serialize, Debug, gltf::derive::Validate, Clone)]
struct RootExtensions {
    #[serde(default, rename = "VRM")]
    vrm: Option<Vrm>,
}

struct GltfExtensions;

impl gltf::json::CustomExtensions for GltfExtensions {
    type Root = RootExtensions;
}

...

let gltf = gltf::Gltf::<GltfExtensions>::from_slice(bytes).unwrap();
let json = gltf.document.as_json();
let extensions = &json.extensions;

if let Some(ext) = extensions.as_ref() {
    if let Some(vrm) = &ext.cuystom.vrm {
         ...
    }
}
```

@3c1u you mentioned wanting VRM support

@alteous I believe this fixes the problems you encountered in https://github.com/gltf-rs/gltf/issues/11#issue-216892693.


expenses:
If this is too big to merge in it's current state then we could do it in 2 parts: the gltf-json changes and then the gltf changes. We'd just have to hardcode it so that instead of `pub struct Document<E: json::CustomExtensions = ()>(json::Root<E>)` we have `pub struct Document(json::Root<()>)`.

expenses:
There are a few things that I need to fix up like documentation. I'd love to get feedback on whether this is the right approach beforehand though.

expenses:
> Here's what adding support for the VRM extension might look like (I'm making a crate for this):

See https://github.com/expenses/mateversum/blob/af068ec23c9cb23a352977422e31b38708acb1ba/vrm-gltf/src/lib.rs

for a more complete implementation.

PR #351, opened 2022-07-12
Fix compilation with `KHR_materials_volume` feature only.

kpreid:
`KHR_materials_volume` needs `Validate`, `texture`, and `extras`, but those items were not imported when exactly that feature was enabled.

To fix this and any other similar problems, and prevent reoccurrences, make the `use` unconditional, with `#[allow(unused_imports)]` to avoid warnings when the relevant features are not enabled.

### DIFF
--- a/gltf-json/src/extensions/material.rs
+++ b/gltf-json/src/extensions/material.rs
@@ -1,16 +1,5 @@
-#[cfg(feature = "KHR_materials_pbrSpecularGlossiness")]
-use crate::material::StrengthFactor;
-#[cfg(any(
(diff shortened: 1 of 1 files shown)


kpreid:
P.S. The extension specification says

> The volume extension needs to be combined with an extension which allows light to transmit through the surface, e.g. `KHR_materials_transmission`.

so one could argue that the `KHR_materials_volume` feature should depend on the `KHR_materials_transmission` feature, instead of adjusting `KHR_materials_volume` to work by itself. But this PR seems to me like a good idea on its own — making the code more robust against future changes.

PR #352, opened 2022-07-12
Implement `Eq`, `Ord`, and `Hash` for `gltf_json::Index`.

kpreid:
These additional trait implementations will allow `Index`es to be stored in maps and compared with others, which will be useful at least for my purposes in generating glTF and testing the generator.

While I was looking at the code, I also eliminated the `unsafe impl`s by changing the phantom type from `PhantomData<*const T>` to `PhantomData<fn() -> T>`, which avoids the `!Send` that phantom `*const T` causes. This is a separate commit and could be split to a separate PR if desired.

### DIFF
--- a/gltf-json/src/root.rs
+++ b/gltf-json/src/root.rs
@@ -20,7 +20,7 @@ pub trait Get<T> {
 }
 
 /// Represents an offset into an array of type `T` owned by the root glTF object.
-pub struct Index<T>(u32, marker::PhantomData<*const T>);
+pub struct Index<T>(u32, marker::PhantomData<fn() -> T>);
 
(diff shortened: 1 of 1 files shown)


kpreid:
@aloucks `PhantomData<T>` would cause it to be `!Send + !Sync` if the `T` was, which doesn't make sense here because the `Index` is just a number and doesn't own a `T`. (It probably doesn't matter at all in this case because `gltf_json`'s data structures will be `Send + Sync`, but it's more correct abstractly.)

`PhantomData<fn() -> T>` expresses “this is a thing which, if you use it correctly, can get you a `T`”. Arguably it should be `-> &T` instead, but then you have lifetime questions (that I don't think would actually change things here).

PR #358, opened 2022-10-01
Fix panic on data slice accesses

repi:
This could easily happen if the binary data for the glTF was too short or empty.

Ideally should return a `Result` and enum error for this rather than just `Option::None` but this is a pretty easy change at least and better than panicking.

Part of #359

### DIFF
--- a/src/accessor/util.rs
+++ b/src/accessor/util.rs
@@ -10,7 +10,9 @@ fn buffer_view_slice<'a, 's>(
 ) -> Option<&'s [u8]> {
     let start = view.offset();
     let end = start + view.length();
-    get_buffer_data(view.buffer()).map(|slice| &slice[start..end])
+    get_buffer_data(view.buffer())
+        .map(|slice| slice.get(start..end))
+        .flatten()
 }
 
 /// General iterator for an accessor.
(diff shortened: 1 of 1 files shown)


src/accessor/util.rs
```diff
@@ -10,7 +10,9 @@ fn buffer_view_slice<'a, 's>(
 ) -> Option<&'s [u8]> {
     let start = view.offset();
     let end = start + view.length();
-    get_buffer_data(view.buffer()).map(|slice| &slice[start..end])
+    get_buffer_data(view.buffer())
+        .map(|slice| slice.get(start..end))
+        .flatten()
```
IcanDivideBy0, line 15:
```
.and_then(|slice| slice.get(start..end))
```

src/accessor/util.rs
```diff
@@ -297,7 +299,7 @@ impl<'a, 's, T: Item> Iter<'s, T> {
                     let start = accessor.offset();
                     let end = start + stride * (accessor.count() - 1) + mem::size_of::<T>();
                     let subslice = if let Some(slice) = buffer_view_slice(view, &get_buffer_data) {
-                        &slice[start..end]
+                        slice.get(start..end)?
                     } else {
                         return None;
                     };
```
IcanDivideBy0, line 305: This can also be simplified:
```rs
let subslice = buffer_view_slice(view, &get_buffer_data)
    .and_then(|slice| slice.get(start..end))?;
```

IcanDivideBy0:
lgtm
`Iter::new`could benefit from a bit of refactoring to use more idiomatic rust and avoid some invatians option unwrap, but that's outside the scope of the PR

IcanDivideBy0:
Can you please either allow edits from maintainers or cherry-pick 749399bcc19daf8cf74cee341c329e182815628c

repi:
thanks! resolved

IcanDivideBy0:
thank you

PR #360, opened 2022-10-23
Fix bounds calculation in export example

Jmgr:
This fixes the bounds calculation in the export example.


IcanDivideBy0:
Thank you!

PR #362, opened 2022-11-05
Add KHR_materials_clearcoat

TomasKralCZ:
Hi !
I added support for the Clearcoat extension.
Note that I mostly adapted the code from the other extensions.


pcwalton:
@alteous Can we get this merged? We're having to work around it in Bevy.

PR #363, opened 2022-11-24
Add transmission ext. support to list of extensions

Entalpi:
Saw that there was a missing supported extension in the README.md list :)


alteous:
Thanks :smile:

PR #373, opened 2023-04-20
Deterministic serialization

alteous:
The data structure for `Primitive::attributes` has been changed from `HashMap` to `BTreeMap` so that serialization is deterministic, i.e., it produces the same output every time. This was not the case with `HashMap` since the hash algorithm uses a randomly generated seed.


alteous:
Closes #336

PR #375, opened 2023-04-25
Syn2, bump msrv to match `image`, and `edition = "2021"`

Elabajaba:
`gltf-derive` still seems to work with the syn2 bump, the only change I had to do was fix it so it'd compile with `edition = 2021`.

The msrv in the readme was wrong, so I bumped it up to match `image`'s current msrv and added rust-version fields to `cargo.toml`.

There were a bunch of clippy errors so I fixed them.


alteous:
Thanks for fixing these!

Issue #380, opened 2023-06-08
Semver violating change in gltf-json v1.2


alice-i-cecile:
This appears to have broken fresh builds of Bevy: see [this CI run](https://github.com/bevyengine/bevy/actions/runs/5210806000/jobs/9402330114?pr=8785) for an example.

Issue #241, opened 2019-07-28
Document semantic versioning guarantees


alteous:
We have three crates under this repository: `gltf`, `gltf-json`, and `gltf-derive`. The main crate is `gltf` and is designed to adhere to [semantic versioning](https://semver.org/) guarantees. The other two crates are considered implementation details and their versions match that of `gltf` for ease of deployment. For diplomatic reasons, `gltf-json` deliberately exists in its own crate although this is not strictly necessary. The fact that `gltf` and `gltf-json` share the same version number _and_ `gltf-json` is re-exported as the `json` module poses a question: does a backward incompatible change in `gltf-json` affect the major version of `gltf`? I think not and it would be worth documenting this so users are not totally surprised.

alteous:
On second thoughts, I'm starting to think we ought to use semantic versioning 'properly' for `gltf-json` and `gltf-derive` at some point. I suggest we start doing so post 1.0 release.

alteous:
I also suggest we remove the `json` module/re-export.

zicklag:
If the `gltf-json` module *is* re-exported in the `gltf` crate, then it *should* affect the major version of `gltf` because it is in the public API. I agree that if we don't want it to affect the major version of `gltf` then it should not be re-exported.

Also, while `gltf-json` is considered an implementation detail of the `gltf` crate, in order to make a glTF exporter, I have to use the `gltf-json` crate, not the `gltf` crate, so it makes sense that `gltf-json` would be its own standalone crate with its own version number that relates to its own public API.

In summary: :+1:

PR #381, opened 2023-06-14
Make byte_offset optional for sparse accessors

elrnv:
This fixes validation of sparse accessors without buffer_views, which must not have a byte_offset. This is not explicit in the spec but is in the official validator (see https://github.com/KhronosGroup/glTF-Validator/issues/207)

### DIFF
--- a/src/accessor/mod.rs
+++ b/src/accessor/mod.rs
@@ -120,9 +120,11 @@ impl<'a> Accessor<'a> {
 
     /// Returns the offset relative to the start of the parent buffer view in bytes.
     ///
-    /// This may be `None` if the corresponding accessor is sparse.
-    pub fn offset(&self) -> Option<usize> {
-        self.json.byte_offset.map(|offset| offset as usize)
+    /// This will be 0 if the corresponding accessor is sparse.
+    pub fn offset(&self) -> usize {
+        // TODO: Change this function to return Option<usize> in the next
+        // version and return None for sparse accessors.
+        self.json.byte_offset.unwrap_or(0) as usize
     }
 
     /// Returns the number of components within the buffer view - not to be confused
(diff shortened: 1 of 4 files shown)


src/accessor/mod.rs
```diff
@@ -119,8 +119,10 @@ impl<'a> Accessor<'a> {
     }
 
     /// Returns the offset relative to the start of the parent buffer view in bytes.
-    pub fn offset(&self) -> usize {
-        self.json.byte_offset as usize
+    ///
+    /// This may be `None` if the corresponding accessor is sparse.
+    pub fn offset(&self) -> Option<usize> {
```
alteous, line 124: This will be a semver breaking change. I'm wondering whether we can return zero instead if it's `None` or whether we should bump the major version. Since https://github.com/gltf-rs/gltf/issues/380 I'm tempted to move to version 2 and start taking semver more seriously as discussed in https://github.com/gltf-rs/gltf/issues/241.

alteous:
Well spotted and thanks for the PR.

src/accessor/mod.rs
elrnv: I think that's fine by me. But this PR already breaks semver for gltf-json. Let me know how you would like to proceed and I'll make the adjustment.

elrnv: @alteous any decision on this? I'm happy to make the change, otherwise feel free to make changes to this PR directly.

alteous: I'm not so concerned about breaking semver for gltf-json. It's not the first time this has happened. As long as the top-level gltf crate is unaffected. I suggest we go with the 'return zero' approach for now.

> Edit: I should add, that if we do return zero, we should probably still change it to return Option<usize> for the next version.

I agree with you. This is something that could be missed easily. I'm not how that could be prevented easily though.

elrnv:
Should be good to go, @alteous.

PR #382, opened 2023-06-17
Added conversion function, with support for 8bits pixel format only

Ithyx:


Ithyx:
Sorry, this was a mistake!

PR #387, opened 2023-09-24
Update link to infographic

attackgoat:
Khronos updated the graphic to include the most recent extensions, removing the old version.

See: https://github.com/KhronosGroup/glTF/pull/2325


alteous:
Thanks :+1:

PR #391, opened 2023-09-25
seek relative offset

Jcskey:


Jcskey:
yes, in some cases, glTF is part of other standards, `seek relative position` can be used without interrupting the reading stream. thanks for your support

PR #395, opened 2023-11-04
Arbitrary Extension API

### LINKED ISSUE #231: Support for custom extensions

CorneliusCornbread:
Fixes #231 

You can access the extensions in the following manner:
```rs
//materials extension, same for all fields
for value in gltf.materials() {
    if let Some(ext) = value.extensions() {
        for values in ext {
            println!("{},{}", values.0, values.1)
        }
    }
}

//root extensions
gltf.extensions();

//or you can use get_extension_value() for a single value instead of the whole map
for value in gltf.materials() {
    println!(
        "{}",
        value.get_extension_value("VRMC_materials_mtoon").unwrap()
    )
}
```

### DIFF
--- a/src/accessor/mod.rs
+++ b/src/accessor/mod.rs
@@ -59,6 +59,8 @@ use crate::{buffer, Document};
 
 pub use json::accessor::ComponentType as DataType;
 pub use json::accessor::Type as Dimensions;
+#[cfg(feature = "extensions")]
+use serde_json::{Map, Value};
 
 /// Utility functions.
 #[cfg(feature = "utils")]
@@ -136,6 +138,22 @@ impl<'a> Accessor<'a> {
     /// Returns the data type of components in the attribute.
     pub fn data_type(&self) -> DataType {
         self.json.component_type.unwrap().0
+    }
+
+    /// Returns the extension values map
+    #[cfg(feature = "extensions")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "extensions")))]
+    pub fn extensions(&self) -> Option<&Map<String, Value>> {
+        let ext = self.json.extensions.as_ref()?;
+        Some(&ext.others)
+    }
+
+    /// Return a value for a given extension name
+    #[cfg(feature = "extensions")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "extensions")))]
+    pub fn get_extension_value(&self, ext_name: &str) -> Option<&Value> {
+        let ext = self.json.extensions.as_ref()?;
+        ext.others.get(ext_name)
     }
 
     /// Optional application specific data.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ byteorder = "1.3"
 gltf-json = { path = "gltf-json", version = "1.3.0" }
 lazy_static = "1"
 urlencoding = { optional = true, version = "2.1" }
+serde_json = { features = ["raw_value"], version = "1.0" }
 
 [dependencies.image]
 default-features = false
@@ -37,6 +38,7 @@ version = "0.24"
 
 [features]
 default = ["import", "utils", "names"]
+extensions = ["gltf-json/extensions"]
 extras = ["gltf-json/extras"]
 names = ["gltf-json/names"]
 utils = []
--- a/gltf-json/Cargo.toml
+++ b/gltf-json/Cargo.toml
@@ -17,6 +17,7 @@ serde_json = { features = ["raw_value"], version = "1.0" }
 [features]
 default = []
 names = []
+extensions = []
 extras = []
 KHR_lights_punctual = []
 KHR_materials_ior = []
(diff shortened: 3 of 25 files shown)


src/accessor/mod.rs
```diff
@@ -138,6 +140,22 @@ impl<'a> Accessor<'a> {
...
+    #[cfg_attr(docsrs, doc(cfg(feature = "extensions")))]
+    pub fn extensions(&self) -> Option<&Map<String, Value>> {
+        let ext = self.json.extensions.as_ref()?;
+        Some(&ext.others)
+    }
+
+    /// Return a value for a given extension name
+    #[cfg(feature = "extensions")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "extensions")))]
+    pub fn get_extension_value(&self, ext_name: &str) -> Option<&Value> {
```
alteous, line 154: Could we rename this simply `extension`? Did you think that was too similar?

alteous:
Thanks for your PR.

Using [`RawValue`](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html) instead of [`Value`](https://docs.rs/serde_json/latest/serde_json/enum.Value.html) would shift deserialisation onto the user but, on the other hand, the user would only pay for the extensions used. I'm curious if you gave this approach a try? I like the proposal as-is nonetheless.

src/accessor/mod.rs
CorneliusCornbread: yeah I felt like the s wasn't dissimilar enough. I could remove the `get` from the name so it'd just be `extension_value()`?

CorneliusCornbread:
> Thanks for your PR.
> 
> Using [`RawValue`](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html) instead of [`Value`](https://docs.rs/serde_json/latest/serde_json/enum.Value.html) would shift deserialisation onto the user but, on the other hand, the user would only pay for the extensions used. I'm curious if you gave this approach a try? I like the proposal as-is nonetheless.

The reason was due to complexity. Raw value requires you to reference it in a data structure as a borrowed value, this causes some problems with some of the attributes on the data structures. I could probably figure it out but I simply went with the regular Value.

Also, the majority of the time, when you are using one extension, you're usually using _all_ the extensions.

If I have some free time this week, I'll make a branch and try implementing this using RawValue

CorneliusCornbread:
> Thanks for your PR.
> 
> Using [`RawValue`](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html) instead of [`Value`](https://docs.rs/serde_json/latest/serde_json/enum.Value.html) would shift deserialisation onto the user but, on the other hand, the user would only pay for the extensions used. I'm curious if you gave this approach a try? I like the proposal as-is nonetheless.

So I looked into it, as far as I can tell I don't believe there's a way we can mix flatten and raw value, it seems flatten will only accept a `Map<String, RawValue>` meaning we'd would have to drop support for the existing supported extensions for arbitrary extension support. If you think that would be preferrable I can rewrite it to do that.

src/accessor/mod.rs
alteous: Sounds good to me. :+1:

alteous:
> So I looked into it, as far as I can tell I don't believe there's a way we can mix flatten and raw value, it seems flatten will only accept a Map<String, Value> meaning we'd would have to drop support for the existing supported extensions for arbitrary extension support. If you think that would be preferrable I can rewrite it to do that.

Thanks for taking a look. If that's the case then it's not worth overhauling everything for it. Let's use the owned `Value` as you proposed originally.

alteous (approved):

PR #396, opened 2023-11-27
Allow handling large external buffers

### LINKED ISSUE #393: Panics on reading a large GLTF due to the .bin being more than ~4GB and `byteLength` using a u32
`byteLength` is internally a u32, but that violates the spec which says there's no hard limit on the size of a buffer, though assets should limit themselves to 2^53 to avoid issues with some json decoders.

https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#buffers-and-buffer-views-overview

```json
	"buffers": [
		{
			"uri": "redacted.bin",
			"byteLength": 17646210956
		}
	],
```

ShaddyDC:
Closes #393 

I've added a new error type for it to make failures explicit, but you could arguably also just use `as usize` and live with potentially weird behaviour when large values are truncated. Presumably, in those cases you'd crash anyway when trying to load the data into memory, though I haven't looked into it more.

I've also silenced some warnings in a separate commit by hiding them behind a `cfg` tag because they annoyed me during development, but I'm happy to remove those changes if you prefer.

### DIFF
--- a/examples/export/main.rs
+++ b/examples/export/main.rs
@@ -71,7 +71,7 @@ fn export(output: Output) {
 
     let (min, max) = bounding_coords(&triangle_vertices);
 
-    let buffer_length = (triangle_vertices.len() * mem::size_of::<Vertex>()) as u32;
+    let buffer_length = (triangle_vertices.len() * mem::size_of::<Vertex>()) as u64;
     let buffer = json::Buffer {
         byte_length: buffer_length,
         extensions: Default::default(),
@@ -87,7 +87,7 @@ fn export(output: Output) {
         buffer: json::Index::new(0),
         byte_length: buffer.byte_length,
         byte_offset: None,
-        byte_stride: Some(mem::size_of::<Vertex>() as u32),
+        byte_stride: Some(mem::size_of::<Vertex>() as u64),
         extensions: Default::default(),
         extras: Default::default(),
         name: None,
@@ -198,7 +198,7 @@ fn export(output: Output) {
                 header: gltf::binary::Header {
                     magic: *b"glTF",
                     version: 2,
-                    length: json_offset + buffer_length,
+                    length: json_offset + buffer_length as u32, // This may truncate long buffers
                 },
                 bin: Some(Cow::Owned(to_padded_byte_vector(triangle_vertices))),
                 json: Cow::Owned(json_string.into_bytes()),
--- a/src/accessor/util.rs
+++ b/src/accessor/util.rs
@@ -8,8 +8,8 @@ fn buffer_view_slice<'a, 's>(
     view: buffer::View<'a>,
     get_buffer_data: &dyn Fn(buffer::Buffer<'a>) -> Option<&'s [u8]>,
 ) -> Option<&'s [u8]> {
-    let start = view.offset();
-    let end = start + view.length();
+    let start = usize::try_from(view.offset()).ok()?;
+    let end = usize::try_from(start as u64 + view.length()).ok()?;
     get_buffer_data(view.buffer()).and_then(|slice| slice.get(start..end))
 }
 
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -4,6 +4,7 @@ use std::ops;
 use crate::Document;
 
 pub use json::buffer::Target;
+#[cfg(feature = "extensions")]
 use serde_json::{Map, Value};
 
 /// A buffer points to binary data representing geometry, animations, or skins.
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -92,8 +92,8 @@ impl<'a> Buffer<'a> {
     }
 
     /// The length of the buffer in bytes.
-    pub fn length(&self) -> usize {
-        self.json.byte_length as usize
+    pub fn length(&self) -> u64 {
+        self.json.byte_length
     }
 
     /// Optional user-defined name for this object.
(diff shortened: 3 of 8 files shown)


src/accessor/util.rs
```diff
@@ -8,8 +8,8 @@ fn buffer_view_slice<'a, 's>(
     view: buffer::View<'a>,
     get_buffer_data: &dyn Fn(buffer::Buffer<'a>) -> Option<&'s [u8]>,
 ) -> Option<&'s [u8]> {
-    let start = view.offset();
-    let end = start + view.length();
+    let start = usize::try_from(view.offset()).ok()?;
+    let end = usize::try_from(start as u64 + view.length()).ok()?;
```
aloucks, line 12: Isn't this going to fail on 32bit systems?

src/buffer.rs
```diff
@@ -91,8 +92,8 @@ impl<'a> Buffer<'a> {
     }
 
     /// The length of the buffer in bytes.
-    pub fn length(&self) -> usize {
-        self.json.byte_length as usize
+    pub fn length(&self) -> u64 {
```
aloucks, line 95: This is just going to lead to casts every time it is used. I think `usize` was the right choice. Buffers exceeding `u32::MAX` length become problematic on 32bit systems. I don't think you can even address a slice in Rust beyond `u32::MAX` on 32bit systems. 

I agree that internally, the length should be stored as `u64` but I think the offsets and lengths should be reported as `usize`.

aloucks: Loading the gltf file should fail on 32bit systems if there are buffers with lengths larger than `u32::MAX`.

ShaddyDC: Hmm, I figured it's worth preserving the actual information losslessly as long as possible so users of the interface can decide how they want to deal with it. Do you want to truncate the information here by casting to usize, which may lead to unexpected behaviour, or should this function be fallible?

src/accessor/util.rs
ShaddyDC: Yes, this will fail on 32bit systems when the buffer is too large to fit into the 32bit address space, which seems fair to me. I guess technically the part of the buffer to be accessed in this function specifically may fit, but I'm not sure that edge case is worth handling. Would you like to handle it differently?

examples/export/main.rs
```diff
@@ -198,7 +198,7 @@ fn export(output: Output) {
                 header: gltf::binary::Header {
                     magic: *b"glTF",
                     version: 2,
-                    length: json_offset + buffer_length,
+                    length: json_offset + buffer_length as u32, // This may truncate long buffers
```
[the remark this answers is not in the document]
ShaddyDC: No, internal glb buffers can only be 32 bits, so the binary header has to stay `u32`. Only external buffers can be larger. For the example, the buffer is fixed and will always fit, so we can just cast using `as`. I left the comment there as a warning for people following the example. If you want, I can elaborate on the comment, unless you want to change the example to use an external buffer instead.

src/buffer.rs
ShaddyDC: @aloucks Do you have an update on how you would prefer this to be handled? It sounds like you'd like to fail when loading a glTF file that has any buffer that is too large and then just hard cast to usize here later? The use case of working with larger files on 32 bit systems is not worth supporting?

aloucks: > Do you want to truncate the information here by casting to usize, which may lead to unexpected behaviour, or should this function be fallible?

No, I was saying that it should fail to load if the buffer length exceeds `usize::MAX` and therefor truncation here could never happen. 

If you want to be able to load the json metadata on 32 bit systems that have buffers that exceed `u32::MAX` length, then I guess reporting it as `u64` is fine. My original thought process was that this is just going to shove the cast onto the user, forcing them them to deal with the error condition, when that condition can't even happen on a 64bit system that could load the file (with buffers).

ShaddyDC: @aloucks I'm sorry to nag you yet again, but would you like me to change this or anything else? I need to handle large files in a project, and while I can of course just use this PR for the crate version, I would prefer to use an official release. It's not very urgent, but it would be nice if we could either merge or shelve this PR this year so I don't have to think about it anymore.

examples/export/main.rs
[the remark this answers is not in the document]
alteous: This is merely an example so I think this is OK. This is a known limitation: https://github.com/KhronosGroup/glTF/issues/2114

src/buffer.rs
alteous: I'm tempted to settle on `usize` as 32 bit systems are the exception rather than the norm nowadays. The [WASM](https://webassembly.org/) runtime however is a popular target for this library and [64 bit addressing](https://webassembly.org/roadmap/#linear-memory-bigger-than-4-gib) is unstable. For this reason, I think it's important the user can handle this scenario gracefully somehow.

For the majority of users running on 64 bit systems, the cast from `u32` to `usize` is already a bit annoying, so I suggest few things:
1. Create a new type `struct Address(u64)` (the name is debatable) in gltf-json.
2. Implement `Validate` for `Address` and report an error on 32 bit systems if it exceeds the `usize` range.
3. In the main/wrapper crate, change the return type for existing `u32` offsets/sizes to `usize`, using `as` to cast internally.

This should cover all of the points mentioned previously. 32 bit systems will still be able to process large glTF (and the wrapper via. [`from_json_without_validation`](https://docs.rs/gltf/latest/gltf/struct.Document.html#method.from_json_without_validation)) or report to the user the out of range size otherwise.

@ShaddyDC, would you like to implement this yourself? I can chip in and push to this branch if you like otherwise.

ShaddyDC: I can take a shot at an implementation, but if you've got the time to do it, I'd appreciate if you could take care of it. I'm not yet familiar with the validation implementation.

alteous: Leave it with me :smile:

alteous: I've opened a PR on your fork: https://github.com/ShaddyDC/gltf/pull/1

ShaddyDC: Lgtm, merged!

ShaddyDC:
@alteous thanks for merging this! When are you planning to do the next release?

alteous:
Likely today or this weekend.

PR #403, opened 2023-12-18
`impl Default for gltf_json::Node`.

kpreid:
Since `Node`s are generally expected to have only a few non-absent fields, providing this `Default` implementation will allow code that is generating `Node`s to be much more concise.


alteous:
The original reason why `Node` didn't have a `Default` implementation was because an empty node is invalid. I think there are better ways of avoiding this though so LGTM. :+1:

Issue #313, opened 2021-11-12
failed to iter over sparse accessor


wsw0108:
sample model: [https://github.com/cx20/gltf-test/tree/master/tutorialModels/SimpleSparseAccessor](https://github.com/cx20/gltf-test/tree/master/tutorialModels/SimpleSparseAccessor)

```rust
fn test() {
    let (document, buffers, images) = gltf::import(file)?;
    let reader = primitive.reader(|buffer| Some(&buffers[buffer.index()]));
    reader.read_positions().unwrap.map(|p| println!("{:?}", p));
}
```

aloucks:
Could you elaborate on what you expected to happen and what actually happened?

Were any positions printed?
Were some, but not all, positions printed? 
Did it panic?

wsw0108:
```
     Running `target\debug\gltf-issue-313.exe ..\m3t\models\tutorialModels\SimpleSparseAccessor\glTF\SimpleSparseAccessor.gltf`
thread 'main' panicked at 'attempt to subtract with overflow', C:\Users\wsw\scoop\persist\rustup\.cargo\registry\src\mirrors.ustc.edu.cn-61ef6e0cd06fb9b8\gltf-0.16.0\src\accessor\util.rs:149:20
```

wsw0108:
No, it does not work.

The `hint` shoule be `accessor.count() - self.counter` according the spec.

derwiath:
We encountered this issue on one of our assets. I dug into it and wrote a fix for it: #411. I added some unit tests based on the repro-code @wsw0108 posted above. Thanks for that 👍 

At the same time I also discovered that the sparse accessor iterator could not read accessors where the `bufferView` field was unset. I took the opportunity to fix this issue as well.

PR #411, opened 2024-01-08
Fix size hint and values read with sparse iterator

derwiath:
In #313 @wsw0108 shows that collecting positions specified using a sparse accessor triggers a panic (`attempt to subtract with overflow`) inside of `size_hint()` of `SparseIter`.

In this PR I have rewritten the code in the issue as two unit tests, one that verifies `size_hint()` and one that verifies that the positions are actually correct. Both tests fail when using latest `main`.

The `SimpleSparseAccessor.gltf` test asset contains 14 positions specified in a sparse accessor where three of them is overrideen by values specified in the sparse section of the accessor. Look at the data layout image here for an overview: [gltf-test/tutorialModels/SimpleSparseAccessor][1].

The two tests pass after delegating the `size_hint()` call to the iterator for the base accessor, it knows how many items are left.

The gltf specification on [Sparse Accessors][2] mentions that a sparse accessor may not have a `bufferView` set
> When accessor.bufferView is undefined, the sparse accessor is initialized as an array of zeros of size (size of the accessor element) * (accessor.count) bytes.

For `SparseIter`, `base` will not be set and `size_hint()` cannot rely on it anymore. The solution is simple, simply pass `accessor.count` to `SparseIter` so that it can use it to know how many items there are. Then it simply deducts `self.counter` from it as that variable keeps track of how many items have been read from the iterator.

The asset `tests/box_sparse.gltf` contains a sparse accessor without a base buffer view. This is used in two unit tests that the behavior is correct, even for these sorts of accessors.

[1]: https://github.com/cx20/gltf-test/tree/master/tutorialModels/SimpleSparseAccessor
[2]: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#sparse-accessors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ The `gltf` crate adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Fixed
+- Fix `attemt to to subtract with overflow`-panic in `size_hint()` of sparse accessor when collecting items.
+- Fix incorrect values returned from `size_hint()` in sparse accessor
+- Add support to read items from sparse accessor without base buffer view
+
 ## [1.4.0] - 2023-12-17
 
 ### Added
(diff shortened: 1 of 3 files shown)


alteous:
Thanks for fixing this!

Issue #409, opened 2024-01-01
2024 / Version 2.0 Plan


alteous:
# 2024 / Version 2.0 Plan

Happy new year all!

As discussed in issue https://github.com/gltf-rs/gltf/issues/385, the next released version is expected to be 2.0 and this version is intended to mark the start of semantic versioning for _all_ the crates in this repository rather than just the top-level `gltf` crate.

There are a few critical requirements that would need to be met for this to be reasonably achievable.

### Merging of validation errors with parsing errors to simplify the JSON data structures

This pertains to the `Checked` type:

 ```rust
pub enum Checked<T> {
    Valid(T),
    Invalid,
}
```

The original idea behind this was to separate validation errors (i.e., "the data does not make sense") from parsing errors (e.g., "an object missing its closing brace"). Where validation errors occur a JSON path to the offending datum can be reported. This is achieved using a custom deserialiser. For example, consider the `camera::Type` enumeration:

```rust
/// Specifies the camera type.
#[derive(Clone, Copy, Debug, Eq, PartialEq)]
pub enum Type {
    /// A perspective projection.
    Perspective = 1,

    /// An orthographic projection.
    Orthographic,
}

impl<'de> de::Deserialize<'de> for Checked<Type> {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: de::Deserializer<'de>,
    {
        struct Visitor;
        impl<'de> de::Visitor<'de> for Visitor {
            type Value = Checked<Type>;

            fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
                write!(f, "any of: {:?}", VALID_CAMERA_TYPES)
            }

            fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
            where
                E: de::Error,
            {
                use self::Type::*;
                use crate::validation::Checked::*;
                Ok(match value {
                    "perspective" => Valid(Perspective),
                    "orthographic" => Valid(Orthographic),
                    _ => Invalid,
                })
            }
        }
        deserializer.deserialize_str(Visitor)
    }
}
```

The use of the `Checked` type allows parsing to complete even if the data is malformed; however, pragmatically speaking, if the data is malformed then what use is it to continue parsing other than to provide the JSON path? Discontinuing this practice would allow the code to be generated trivially by `serde_derive` instead. For exporters, it would also simplify populating the JSON data structures.
    
```rust
/// Specifies the camera type.
[...]

Issue #400, opened 2023-12-15
Fails to load gltfPack-0.20 output


kvark:
> thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `12`,
 right: `6`', C:\Users\kvark\.cargo\registry\src\index.crates.io-6f17d22bba15001f\gltf-1.3.0\src\accessor\util.rs:348:17note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

This is happening on "reader.read_positions()".

Example file:
[Fostral-shape.zip](https://github.com/gltf-rs/gltf/files/13681869/Fostral-shape.zip)

I suspect this is because gltfPack-0.20 outputs 16-bit integer vertex coordinates by default instead of 32-bit floating points.

alteous:
Thanks for reporting. This model is using [`KHR_mesh_quantization`](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_mesh_quantization/README.md) which isn't handled by the reader utility.

The extension is marked as required in the example file, so the library ought report it as unsupported in some way. It begs the question: should the library reject any glTF that has unsupported extensions under `extensionsRequired`? In the next release there is an [arbitrary extension API](https://github.com/gltf-rs/gltf/pull/395), so technically the burden would be passed onto the user.

A short term solution might be to fail creating the `Reader` for assets with extensions required matching `*_mesh_*`. Obviously, the long-term solution would be to add support for this extension.

alteous:
https://github.com/gltf-rs/gltf/pull/418 will now report an error if the crate can't handle a particular extension, including this one, so I've rebranded this issue as a feature request to add support for this feature.

sotrh:
Is there no way to bypass the error if I need to use quantization @alteous ?

sotrh:
I created a PR for this #428

PR #392, opened 2023-10-14
Support KHR_texture_basisu extension

aW4KeNiNG:
Initial support for the KHR_texture_basisu extension.
Extension information: https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_texture_basisu/README.md

Issue #385, opened 2023-08-27
gltf-json 1.3 breaks semver


Ralith:
https://github.com/gltf-rs/gltf/pull/381 changed the type of a public field without bumping the major version number of gltf-json. This broke my code.

kpreid:
The same change also broke my glTF export code. I'm aware that this is an intentional choice regarding the versioning of `gltf-json`, but I'd like to ask you to reconsider it in the future. As far as I'm aware, the only way to _generate_ glTF files (without writing a set of serde structure types from scratch) is to depend directly on `gltf-json` and use its structure fields.

kpreid:
And with the latest release I'm reminded of more ways this is a problem: if something in the dependency graph starts to depend on a newer `gltf-json`, and `gltf` is also in the graph but nothing depends on the latest version, then Cargo will update `gltf-json` without `gltf`, producing a broken build.

Please start using accurate semver for `gltf-json` — any time a type is changed incompatibly etc., its *major* version must be bumped. (Yes, this means it must go out of sync with `gltf`. But at least until `gltf 2.0.0` comes along, you can have `gltf 1.5` paired with `gltf-json 5.0.0` and so on.)

<https://crates.io/crates/cargo-semver-checks> can help validate that the version is properly incremented when changes occur.

alteous:
Okay I see this situation is clearly irritating lots of people. I will make the next release 2.0, ensure the semantic versioning applies to the underlying data structures thereon. This will likely entail merging gltf-json with the main crate. cargo-semver-checks is a good suggestion.

PR #413, opened 2024-02-18
Fix noncomformant behavior by not requiring the `source` field to be present in textures.

pcwalton:
It's perfectly legal for `source` to not be present. Typically, this will happen when an extension such as `KHR_texture_basisu` provides the actual source.

The glTF spec § 5.29 states unambiguously that `source` is an optional field [1]. The validator backs this up (see the `blank_texture.gltf` test case, which validates). Furthermore, the `KHR_texture_basisu` extension provides the following example in "Using without a fallback" [2]:

```json
"textures": [
{
    "extensions": {
	"KHR_texture_basisu": {
	    "source": 0
	}
    }
}
],
```

This doesn't contain a `source` field. Parsing this today with the `gltf` crate causes a panic in `serde`.

I ran into this with the `gltf-transform` [3] tool, which follows the standard when compressing textures and therefore produces valid glTF files that the `gltf` crate panics when attempting to load.

[1]: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-texture

[2]: https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_texture_basisu/README.md#using-without-a-fallback

[3]: https://github.com/donmccurdy/glTF-Transform

### DIFF
--- a/gltf-json/src/texture.rs
+++ b/gltf-json/src/texture.rs
@@ -179,7 +179,8 @@ pub struct Texture {
     pub sampler: Option<Index<Sampler>>,
 
     /// The index of the image used by this texture.
-    pub source: Index<image::Image>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub source: Option<Index<image::Image>>,
 
     /// Extension specific data.
     #[serde(default, skip_serializing_if = "Option::is_none")]
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -157,12 +157,12 @@ impl<'a> Texture<'a> {
             .unwrap_or_else(|| Sampler::default(self.document))
     }
 
-    /// Returns the image used by this texture.
-    pub fn source(&self) -> image::Image<'a> {
-        self.document
-            .images()
-            .nth(self.json.source.value())
-            .unwrap()
+    /// Returns the image used by this texture, if any.
+    pub fn source(&self) -> Option<image::Image<'a>> {
+        self.json
+            .source
+            .as_ref()
+            .map(|index| self.document.images().nth(index.value()).unwrap())
     }
 
     /// Returns extension data unknown to this crate version.
--- a/tests/blank_texture.gltf
+++ b/tests/blank_texture.gltf
@@ -1,0 +1,8 @@
+{
+  "asset": {
+    "version": "2.0"
+  },
+  "textures": [
+    {}
+  ]
+}
--- a/tests/import_sanity_check.rs
+++ b/tests/import_sanity_check.rs
@@ -60,7 +60,8 @@ fn run() -> Result<(), Box<dyn StdError>> {
         }
     }
 
-    sparse_accessor_without_buffer_view_test()
+    sparse_accessor_without_buffer_view_test()?;
+    blank_texture_test()
 }
 
 /// Test a file with a sparse accessor with no buffer view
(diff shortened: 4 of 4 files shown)


alteous:
Thanks for the PR, this one is troubling...

The crate doesn't support the `KHR_texture_basisu` extension explicitly, so in a sense it's correct to reject the file. If it did load as an `Option` type then the application would have to use the [arbitrary extension API](https://github.com/gltf-rs/gltf/pull/395) to avoid the ['undefined behaviour'](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-texture) case:

> When undefined, an extension or other mechanism **SHOULD** supply an alternate texture source, otherwise behavior is undefined.

This is a similar situation to https://github.com/gltf-rs/gltf/issues/400 where the crate fails because an extension changes the way the glTF file is interpreted. Forcing the application to check potentially many different extensions for the 'real' texture data source is undesirable at best.

Changing the type to `Option` at this point would also break semver and I've promised not to do that until the next major release—see #409 and #385.

I think the least intrusive way of dealing with this would be to reject any glTF files containing _required_ extensions that the library does not support explicitly. This should prevent situations like this and #400 from occurring whilst leaving optional extensions available via. the arbitrary extension API.

> Parsing this today with the gltf crate causes a panic in serde.

Nitpicking: it doesn't panic, it produces an error, but I understood what you meant.
```
cargo run --example gltf-display blank-texture.gltf
     Running `target/debug/examples/gltf-display blank-texture.gltf`
thread 'main' panicked at examples/display/main.rs:16:20:
runtime error: Deserialize(Error("missing field `source`", line: 5, column: 18))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

teddemunnik:
>The crate doesn't support the KHR_texture_basisu extension explicitly, so in a sense it's correct to reject the file

There has been a PR open for a while implementing such support. (https://github.com/gltf-rs/gltf/pull/392)

>This is a similar situation to https://github.com/gltf-rs/gltf/issues/400 where the crate fails because an extension changes the way the glTF file is interpreted.

The extension `KHR_texture_basisu` doesn't change the base spec. `texture.source`  is not a required field. It is recommended that a fallback mechanism is available when it's left out, but again that's not a hard requirement in the spec.

I completely understand and appreciate the concern about breaking semver though. Is there a development branch for the next semver breaking release that changes like this could be added to?

Perhaps it could be good to in the future make users of the library specify the different extensions they support on the gltf-json level, so that if they don't specify an extension files which mark it as required may fail to load. The higher level gltf API may already specify the extensions it supports such as texture formats and then hide this from the user.

alteous:
> Is there a development branch for the next semver breaking release that changes like this could be added to?

The closest branch at the moment would be https://github.com/alteous/gltf/tree/v2-with-extension-macro. I intend to make a PR for this soon.

https://github.com/gltf-rs/gltf/pull/418 will report a validation error where required extensions are not handled by the wrapper crate. It makes the texture source optional too. This should work for #400 too.

alteous:
I've amended #418 to use a sentinel index value rather than `Option`. This way should work for everyone: it won't break builds, it'll solve the required extension problem, and the new feature allows power users to handle the empty texture case.

alteous:
This should be covered by https://github.com/gltf-rs/gltf/pull/418 now. Please reopen otherwise. :smile:

PR #425, opened 2024-06-19
Add support for `EXT_texture_webp` v2

marstaik:
I was recently working in [`three_d` ](https://github.com/asny/three-d) which uses gltf in it's asset parser. I realized that neither `three_d` nor `gltf` actually had support for WebP.

I noticed some unmerged PR's and decided to update them in hopes of getting this in (and a dependent PR in `three_d` merged as well).

Going down the rabbit hole, I noticed some gremlin interaction with the `allow_empty_textures` feature.

* Validate needs to also acknowledge webp's potential existence
* Don't want to hide things from the `gltf_json::Texture` as it is the source of truth (ie make source private and provide a helper `source()`)

I took the approach of making the `source()` function in texture use a `primary_source()` function, which prefers the `webp` and falls back to whatever was there originally.

However, I believe the most accurate solution is to permanently enable `allow_empty_textures`, and remove that flag, and make applications handle `webp` images (if they have the `EXT_texture_webp` feature enabled).
Perhaps `allow_empty_texture` in its off-state is not gltf compliant anyways?

Original PR: https://github.com/gltf-rs/gltf/pull/365
Related: https://github.com/gltf-rs/gltf/pull/413


alteous:
Thanks!

PR #442, opened 2024-12-17
implement PartialEq on Root and its attributes

rydb:
I am trying to use this library to serialize meshes for my project into the gltf format. In order to confirm I'm doing it correctly, I would like to compare a mesh gltf imported from a working exporter, and the mesh gltf exported from the one I'm working on.

However, I cannot compare gltf files as `Root` does not currently implement `PartialEq`

This pr resolves this by implementing PartialEq on `Root` and all attributes that `Root` uses.

### DIFF
--- a/gltf-json/src/accessor.rs
+++ b/gltf-json/src/accessor.rs
@@ -83,7 +83,7 @@ pub mod sparse {
     use crate::extensions;
 
     /// Indices of those attributes that deviate from their initialization value.
-    #[derive(Clone, Debug, Deserialize, Serialize, Validate)]
+    #[derive(Clone, Debug, Deserialize, Serialize, Validate, PartialEq)]
     pub struct Indices {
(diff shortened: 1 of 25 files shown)


rydb:
tried to test the check failures locally with act and I'm getting these two errors
```
[Tests/Run test]   ❗  ::error::Input required and not supplied: token
[Tests/Run test]   ❌  Failure - Main Checkout glTF Sample Assets
```
not sure how to fix this 🤷

rydb:
> tried to test the check failures locally with act and I'm getting these two errors
> 
> ```
> [Tests/Run test]   ❗  ::error::Input required and not supplied: token
> [Tests/Run test]   ❌  Failure - Main Checkout glTF Sample Assets
> ```
> 
> not sure how to fix this.

gave test suite auth token and this bug went away. Now waiting on:
https://github.com/serde-rs/json/pull/1227

to be merged to finish this pr.

PR #444, opened 2025-01-12
examples/export: Replace unsound `to_padded_byte_vector()` implementation with `bytemuck`.

kpreid:
The function `to_padded_byte_vector()` is unsound because:

* It accepts an arbitrary `Vec<T>` without checking that `T` contains no padding, which is UB to read in any way including by reinterpreting as `u8`s.
* It produces a `Vec` which thinks it has a different alignment than the allocation was actually created with.

To fix these problems, this change:

* Uses `bytemuck` to check the no-padding condition.
* Creates a new `Vec` instead of trying to reuse the existing one. (Conditional reuse would be possible, but more complex.)

An alternative to using `bytemuck` would be to make `to_padded_byte_vector()` an `unsafe fn` (or to accept `Vertex` only instead of `T`). However, I think it is valuable to demonstrate how to do this conversion using safe tools, to encourage people to use safe code instead of writing unsafe code without fully understanding the requirements.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ members = ["gltf-derive", "gltf-json"]
 
 [dev-dependencies]
 approx = "0.5"
+bytemuck = { version = "1.21.0", features = ["derive"] }
 
 [dependencies]
 base64 = { optional = true, version = "0.13" }
--- a/examples/export/main.rs
+++ b/examples/export/main.rs
@@ -16,7 +16,7 @@ enum Output {
     Binary,
 }
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, bytemuck::NoUninit)]
 #[repr(C)]
 struct Vertex {
     position: [f32; 3],
@@ -42,15 +42,14 @@ fn align_to_multiple_of_four(n: &mut usize) {
     *n = (*n + 3) & !3;
 }
 
-fn to_padded_byte_vector<T>(vec: Vec<T>) -> Vec<u8> {
-    let byte_length = vec.len() * mem::size_of::<T>();
-    let byte_capacity = vec.capacity() * mem::size_of::<T>();
-    let alloc = vec.into_boxed_slice();
-    let ptr = Box::<[T]>::into_raw(alloc) as *mut u8;
-    let mut new_vec = unsafe { Vec::from_raw_parts(ptr, byte_length, byte_capacity) };
+fn to_padded_byte_vector<T: bytemuck::NoUninit>(data: &[T]) -> Vec<u8> {
+    let byte_slice: &[u8] = bytemuck::cast_slice(data);
+    let mut new_vec: Vec<u8> = byte_slice.to_owned();
+
     while new_vec.len() % 4 != 0 {
         new_vec.push(0); // pad to multiple of four bytes
     }
+
     new_vec
 }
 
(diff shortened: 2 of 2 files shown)


kpreid:
The CI failure seems to be in an unrelated import test.

alteous:
I'm surprised this is unsound but I trust your intuition. I'd like to understand better:

> It accepts an arbitrary `Vec<T>` without checking that `T` contains no padding, which is UB to read in any way including by reinterpreting as `u8`s.

Why is this undefined behaviour? Padding bytes between `T`s would be acceptable in the output. 

> It produces a Vec which thinks it has a different alignment than the allocation was actually created with.

This is an interesting point. In C, I am used to expecting `malloc` to produce a region that is suitably aligned for any type. Rust's [`Allocator` trait](https://doc.rust-lang.org/stable/std/alloc/trait.Allocator.html#memory-fitting) docs say the "block must be allocated with the same alignment as [layout.align()](https://doc.rust-lang.org/stable/std/alloc/struct.Layout.html#method.align)", which I think is the problem you're referring to because `align_of::<u8>()` ≤ `align_of::<T>()` in general and this could be a problem for reallocation and deallocation.

kpreid:
> Why is this undefined behaviour?

Excerpted from <https://doc.rust-lang.org/reference/behavior-considered-undefined.html>:

> Rust code is incorrect if it exhibits any of the behaviors in the following list.
> 
> * Producing an [invalid value](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#invalid-values).
>
> ... Whether a value is valid depends on the type:
>
> * An integer (i\*/u\*), floating point value (f\*), or raw pointer must be initialized, i.e., must not be obtained from [uninitialized memory](http://llvm.org/docs/LangRef.html#undefined-values).

Reasons why the *design* is such that this is undefined behavior include that the compiler is permitted to do things like

* transform a given set of reads of a place, that is not being mutated between any of them, into a different number and timing of reads, *and*
* not actually store the bytes corresponding to padding (e.g. if the struct is entirely stored in registers, or was copied field-by-field to a different location).

Both of these transformations are critical to being able to produce efficient code, and just these two applied together could mean that a value that was read from uninitialized memory appears to have an unstable, changing value, because it ended up being read multiple times from _different_ pieces of uninitialized memory.

Of course, it’s _unlikely_ that any of these specific bad effects would apply to a simple bulk copy as this code is doing, but there isn’t and can’t be an exception in the rules of UB for “that would be silly”. Rigor is necessary to be able to have sound, highly optimizing compilers.

It’s also permitted for the _hardware_ to trap on reads of uninitialized memory (though in most cases today, this only happens at page granularity).

---

> Padding bytes between `T`s would be acceptable in the output.

Separately from UB considerations, I disagree that they are acceptable. They may not affect the behavior of a glTF loader, but there are many other reasons to not to write them:

* Copying the contents of uninitialized memory to external destinations can lead to leaking secrets from previous uses of the same memory (as in the famous bug [Heartbleed](https://en.wikipedia.org/wiki/Heartbleed)).
* In use cases like asset pre-processing, it’s useful for the process to be deterministic, i.e. the output is the same if the input is the same. Padding bytes would not be reliably the same.
* Including arbitrary bytes makes the data less compressible.

It’s fine to include _guaranteed zero_ bytes, of course, but not uninitialized bytes.

-----

> I am used to expecting `malloc` to produce a region that is suitably aligned for any type.

Indeed that is a difference, but the key difference here is that Rust deallocation has to be given the layout information, whereas `free()` is obligated to keep track of any information it needs itself.

> Rust's [Allocator trait](https://doc.rust-lang.org/stable/std/alloc/trait.Allocator.html#memory-fitting) docs say the "block must be allocated with the same alignment as [layout.align()](https://doc.rust-lang.org/stable/std/alloc/struct.Layout.html#method.align)", which I think is the problem you're referring to because align_of::<u8>() ≤ align_of::<T>() in general and this could be a problem for reallocation and deallocation.

This is *true*, but not a good way to think about writing unsafe code. When you call an unsafe function, you must obey all of the safety requirements from **that specific function’s documentation**. Those in [`Vec::from_raw_parts()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts) include:

> * `T` needs to have the same alignment as what `ptr` was allocated with.

This rule was violated, so no further analysis is required to conclude that the call is unsound.

alteous:
Thanks for your detailed explanations, I agree with your points.

PR #448, opened 2025-03-19
Add `no_std` Support to `gltf` and `gltf-json`

bushrat011899:
# Objective

Adding `no_std` support is generally desirable across the Rust ecosystem, allowing more users to consume and contribute to the project.

## Solution

- Added `std` feature to `gltf-json` and `gltf` (enabled by default). Note that when disabled the MSRV increases.
- Added `spin`, and `libm` features to `gltf` for compatibility on `no_std` targets.
- Added a CI task to catch regressions in `no_std` support. Note that since Cargo has no syntax for "all features _except_ ...", this must list all features except those which explicitly require `std`.

---

## Notes

- I'm mainly a contributor to the Bevy game engine, where I have been progressively adding `no_std` support. Adding such support to `gltf` would be desirable but is not immediately required.
- This is my first time contributing to this project. Please let me know if there's anything I can do to assist with this PR!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,3 +34,19 @@ jobs:
 
       - name: Semantic versioning checks
         uses: obi1kenobi/cargo-semver-checks-action@v2
+  
+  check-compiles-no-std:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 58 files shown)


barabadzhi:
Hey, any chance we can revive the `no_std` initiative?

PR #449, opened 2025-03-19
allow empty accessors

### LINKED ISSUE #346: `Accessor` implementation not conformant with specification
1. The `Validate` implementation asserts `buffer_view` and `sparse`
must not both be empty, which is actually [allowed](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#_accessor_bufferview).

2. The `Accessor` always has `byte_offset` even when `buffer_view`
is none, which is [not allowed](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#_accessor_byteoffset).

robtfm:
fixes #346

accessors without `bufferView` or `sparse` should be treated as pure zeros.

### DIFF
--- a/gltf-json/src/accessor.rs
+++ b/gltf-json/src/accessor.rs
@@ -1,5 +1,5 @@
-use crate::validation::{Checked, Error, USize64};
-use crate::{buffer, extensions, Extras, Index, Path, Root};
+use crate::validation::{Checked, USize64};
+use crate::{buffer, extensions, Extras, Index};
 use gltf_derive::Validate;
 use serde::{de, ser};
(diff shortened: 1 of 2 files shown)


alice-i-cecile:
Thanks :) From my understanding of the GLTF spec this looks like a reasonable graceful failure mechanism.

MalekiRe:
This fix is critical to using a subset of gLTFs exported from the Godot engine ( VRM models )

robtfm:
is there anything more i can do? the requested changes are done and it's quite a simple scoped change.

PR #458, opened 2025-07-14
feat: read custom mesh attribute in user defined extension

windwhiterain:
I add new function "read_extension" for Reader in mesh mod that comsume a type var T and a path to the accessor index in the json value of extensions, then use the accessor of given index, interpret it as Iterator of T value, return None if no such path exists, panic if the accessor is not exists or can't made into Iterator of T.

### DIFF
--- a/src/mesh/mod.rs
+++ b/src/mesh/mod.rs
@@ -56,6 +56,9 @@ pub mod iter;
 #[cfg_attr(docsrs, doc(cfg(feature = "utils")))]
 pub mod util;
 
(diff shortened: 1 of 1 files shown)


alteous:
Hi, could you run `cargo fmt` on this and provide some example code (in comments is OK)?

windwhiterain:
> Hi, could you run `cargo fmt` on this and provide some example code (in comments is OK)?

Hi, thanks for your reply, i will do this in this weekends.

PR #463, opened 2025-11-24
Add EXT_meshopt_compression and KHR_mesh_quantization support

kikijiki:


jaynewey:
Just to note: `KHR_mesh_quantization` doesn't seem to work when `KHR_texture_transform` is also applied

kikijiki:
@jaynewey Thanks for catching that. The library wasn't tracking the normalized flag for U8/U16 texture coordinates, it always normalized them. I fixed that and added a couple of small tests.

Issue #349, opened 2022-06-28
Fails to parse gltf with scenes without nodes


FredrikNoren:
This file was exported from blender, but fails to parse with this library:

[BuildingGenerator2HardNormalsGLTF.gltf.zip](https://github.com/gltf-rs/gltf/files/9001831/BuildingGenerator2HardNormalsGLTF.gltf.zip)

It seems like it's because it's scenes look like this:

```
   "scenes" : [
        {
            "name" : "Scene",
            "nodes" : [
                0
            ]
        },
        {
            "name" : "Scene.001"
        },
        {
            "name" : "Scene.002"
        }
    ],
```

PR #466, opened 2026-02-13
Allow missing nodes field in scenes

unlessgames:
It's a tiny fix for #349 

As per the [GLTF spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#nodes-and-hierarchy) 
>glTF assets MAY define nodes

Serialization already omits the `nodes` field if the list is empty and deserialization shouldn't throw an error for a missing `nodes` field either.


alteous:
Thanks! :smile:

PR #467, opened 2026-02-16
feat: add KHR_texture_basisu extension (JSON-level only)

mholiv:
Adds a KHR_texture_basisu feature flag that lets textures reference Basis Universal (KTX2) compressed images via the extension's `source` field. Follows the same pattern as EXT_texture_webp: a TextureBasisu struct, a check in primary_source(), and "image/ktx2" in the valid MIME types list.

Based on #392 by @aW4KeNiNG. The original PR had drifted pretty far behind main over time, so I rebased the basisu bits onto current main.


ambrosia13:
@alteous is there any update on this? I've been trying to load Amazon Bistro from [here](https://github.com/qian-o/GLTF-Assets/tree/main/Bistro), but it's panicking by lack of support for this extension

PR #476, opened 2026-08-06
Bump base64 to 0.23.1

martinfrances107:
Also fix deprecated function 

for example 

```
-            Scheme::Data(_, base64) => base64::decode(base64).map_err(Error::Base64),
+            Scheme::Data(_, base64) => STANDARD.decode(base64).map_err(Error::Base64),
```


martinfrances107:
Error are internal test runner flaws

